PR #1, opened 2020-04-15
Port multicluster secrets from Service Mesh Hub

joekelley:


pkg/multicluster/access/kubeconfig_secret.go
```diff
@@ -0,0 +1,97 @@
...
+// SecretToConfigConverterProvider returns an implemented SecretToConfigConverter
+func SecretToConfigConverterProvider() SecretToConfigConverter {
+	return SecretToConfig
+}
+
+// SecretToKubeConfig is an implementation of SecretToConfigConverter.
+func SecretToConfig(secret *kubev1.Secret) (clusterName string, config *Config, err error) {
+	if len(secret.Data) > 1 {
+		return "", nil, SecretHasMultipleKeys(secret.ObjectMeta)
+	}
```
joekelley, line 73: This is the only place we enforce that there be only one key per secret, and we allow writing multiple kubeconfigs in a secret in a function above. Curious whether others think we should allow multiple kubeconfigs per secret.

I don't feel too strongly, but if we do introduce a cluster CRD, it would simplify things a bit if we could count on a 1-1 relationship between cluster CRDs and kubeconfig secrets.

pkg/multicluster/access/kubeconfig_secret.go
```diff
@@ -0,0 +1,97 @@
...
+import (
+	kubev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// KubeConfigSecretLabel is applied to a kubernetes secret to indicate that the secret contains a kubeconfig.
+const KubeConfigSecretLabel = "solo.io/kubeconfig"
```
ilackarms, line 12: i'm not sure about using a proprietary secret label here

pkg/multicluster/access/kubeconfig_secret.go
```diff
@@ -0,0 +1,97 @@
...
+type KubeConfig struct {
+	// the actual KubeConfig
+	Config api.Config
+	// expected to be used as an identifier string for a cluster
+	// stored as the key for the KubeConfig data in a kubernetes secret
+	Cluster string
+}
+
+// KubeConfigToSecret converts a single KubeConfig to a secret with the provided name and namespace.
+func KubeConfigToSecret(name string, namespace string, kc *KubeConfig) (*kubev1.Secret, error) {
```
ilackarms, line 25: are both methods necessary? i wonder if we could live with just one

pkg/multicluster/access/kubeconfig_secret.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+		secretData[kc.Cluster] = rawKubeConfig
+	}
+	return &kubev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels:    map[string]string{KubeConfigSecretLabel: "true"},
+			Name:      name,
+			Namespace: namespace,
+		},
+		Type: kubev1.SecretTypeOpaque,
```
ilackarms, line 48: have we considered using `KubeConfigSecretLabel` as the `Type` instead of Opaque? I think this is the intended convention for the Type field

pkg/multicluster/access/kubeconfig_secret.go
```diff
@@ -0,0 +1,97 @@
...
+	ApiConfig    *api.Config
+	RestConfig   *rest.Config
+}
+
+// SecretToConfigConverter functions extract the cluster name and *Config from a KubeConfig secret.
+// If the provided secret is not a KubeConfig secret, an error is returned.
+type SecretToConfigConverter func(secret *kubev1.Secret) (clusterName string, config *Config, err error)
+
+// SecretToConfigConverterProvider returns an implemented SecretToConfigConverter
+func SecretToConfigConverterProvider() SecretToConfigConverter {
```
ilackarms, line 65: do we need these few lines here? i believe they're used for wire; i don't think we should be opinionated about how DI is going to work in projects

pkg/multicluster/access/kubeconfig_secret.go
ilackarms: also want to understand the reasoning here. is this a real constraint? are there situations where multiple kubeconfigs/clusters would be written to the same secret?

pkg/multicluster/access/kubeconfig_secret.go
joekelley: removed the function that took multiple kubeconfigs for now, happy to restore it if we decide to support multiple kubeconfigs per secret.

pkg/multicluster/access/errors.go
```diff
@@ -0,0 +1,35 @@
+package access
```
ilackarms, line 1: what do you think of the name `kubeconfig` for this package? considering that we are exporting functions like `KubeConfigToSecret`, `access` seems a bit vague. wdyt?

pkg/multicluster/access/kubeconfig_secret_test.go
```diff
@@ -0,0 +1,66 @@
...
+    cluster: test-name
+    user: test-name
+  name: test-name
+current-context: test-name
+kind: Config
+preferences: {}
+users:
+- name: test-name
+  user:
+    token: alphanumericgarbage
```
ilackarms, line 35: lol

joekelley: pulled from service mesh hub, feels like a bit of @GrahamGoudeau test data to me : )

pkg/multicluster/access/errors.go
joekelley: i was trying to be clever / avoid future references like "kubeconfig.KubeConfigToSecret", and maybe account for non-kubernetes client configs, but we can cross that bridge when we get to it. will rename for clarity.

pkg/multicluster/access/kubeconfig_secret.go
joekelley: service mesh hub team is moving toward one per secret, seems good to me.

pkg/multicluster/access/kubeconfig_secret.go
joekelley: service mesh hub team is moving toward one per secret, seems good to me.

joekelley: longer term, might make sense to add a validating webhook to block kubeconfig-type secrets with multiple data entries

pkg/multicluster/kubeconfig/kubeconfig_secret.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	kubev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/tools/clientcmd/api"
+)
+
+// KubeConfigSecretType is used to indicate which kubernetes secrets contain a KubeConfig.
+const KubeConfigSecretType = "solo.io/kubeconfig"
```
EItanya, line 12: nit: rename `SecretType` since the package is kubeconfig anyway

joekelley:
when this merges i'll make the Code Gen, Tests, and changelog-bot checks mandatory

PR #4, opened 2020-04-20
multicluster watcher

joekelley:
TODO:
- [x] expand e2e coverage to include a case where a cluster is removed (and added again?)
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/16


generate.go
```diff
@@ -0,0 +1,56 @@
...
+	// custom proto imports
+	protoImports.External["k8s.io/api"] = []string{
+		"core/v1/*.proto",
+	}
+	protoImports.External["k8s.io/apimachinery"] = []string{
+		"pkg/**/*.proto",
+	}
+
+	cmd := &codegen.Command{
+		AppName:         "dev-portal",
```
ilackarms, line 29: `dev-portal`

generate.go
```diff
@@ -0,0 +1,56 @@
...
+		"pkg/**/*.proto",
+	})
+
+	// custom proto imports
+	protoImports.External["k8s.io/api"] = []string{
+		"core/v1/*.proto",
+	}
+	protoImports.External["k8s.io/apimachinery"] = []string{
+		"pkg/**/*.proto",
+	}
```
ilackarms, line 26: we can eliminate these proto imports

generate.go
```diff
@@ -0,0 +1,56 @@
...
+					Version: "v1",
+				},
+				Module: "k8s.io/api",
+				Resources: []model.Resource{
+					{
+						Kind: "Secret",
+					},
+					{
+						Kind: "ConfigMap",
+					},
```
ilackarms, line 44: we can remove configmap

generate.go
```diff
@@ -0,0 +1,56 @@
...
+				},
+				RenderClients:         true,
```
ilackarms, line 46: i do not believe we need clients (controller should be enough)

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			errs.Errors = append(errs.Errors, err)
+		}
+	}
+
+	return errs.ErrorOrNil()
+}
+
+func (c *clusterWatcher) ReconcileSecretDeletion(req reconcile.Request) {
+	// TODO we have to enforce that the cluster name is the resource name
```
joekelley, line 101: https://github.com/solo-io/skv2/pull/2

pkg/multicluster/clientset.go
```diff
@@ -0,0 +1,81 @@
...
+	"context"
+	"sync"
+
+	"github.com/rotisserie/eris"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// ClientSet provides access to a client.Client for any registered cluster.
+// Only one ClientSet should be created within a multicluster system.
```
ilackarms, line 13: nit: in Kubernetes, a `clientset` usually refers to a set of REST clients for an API group.

i would suggest `Client` here, wdyt?

pkg/multicluster/clientset.go
```diff
@@ -0,0 +1,81 @@
...
+type ClientSet interface {
+	// Cluster returns a client.Client for the given cluster if one is available, else errors.
+	Cluster(name string) (client.Client, error)
+	getManager(cluster string) (manager.Manager, error)
+	setManager(cluster string, mgr manager.Manager, cancel context.CancelFunc)
+	deleteManager(cluster string)
```
ilackarms, line 19: what is the purpose of requiring private methods on a public interface?

pkg/multicluster/kubeconfig/predicate.go
```diff
@@ -0,0 +1,36 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+var _ predicate.Predicate = SecretPredicate{}
+
+// SecretPredicate is a controller-runtime predicate that filters for KubeConfig secrets.
+type SecretPredicate struct{}
```
ilackarms, line 13: does this need to be public?

pkg/multicluster/clientset.go
joekelley: The private methods prevent users from implementing the interface. The implementation in this package is used by the ClusterWatcher to store the canonical set of managers per cluster. Thus, any other implementation of the interface that just exposes the `Cluster` function would not be compatible with the watcher.

This approach was inspired by https://github.com/golang/lint/issues/210#issuecomment-245736660.

Alternatively, I could have just used a struct with some private/public functions. That would be simpler.

pkg/multicluster/kubeconfig/predicate.go
joekelley: nope

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+const (
+	LocalCluster = ""
+)
+
+// AddClusterHandler is called when a new cluster is added.
+// The provided context is cancelled when a cluster is removed, so any teardown behavior for removed clusters
+// should take place when ctx is cancelled.
+type AddClusterHandler func(ctx context.Context, cluster string, mgr manager.Manager) error
```
ilackarms, line 24: i'd prefer to use an interface here for easier mocking and portability, wdyt?

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,111 @@
...
+
+// ClusterWatcher calls ClusterHandlers and maintains a setManager of active cluster credentials.
+type ClusterWatcher interface {
+	controller.SecretReconciler
```
ilackarms, line 28: what is the purpose of this interface? It's not clear to me why we would define an interface on top of SecretReconciler

joekelley: this is an artifact from a previous implementation, removing

pkg/multicluster/clientset.go
joekelley: In the latest iteration, I split this interface into two:

```go
type ManagerSet interface {
	getManager(cluster string) (manager.Manager, error)
	setManager(cluster string, mgr manager.Manager, cancel context.CancelFunc)
	deleteManager(cluster string)
}
```
and
```go
type Client interface {
	// Cluster returns a client.Client for the given cluster if one is available, else errors.
	Cluster(name string) (client.Client, error)
}
```

Client and the clusterWatcher both depend on ManagerSet. Client uses it in order to build clients from an existing cluster manager, and the clusterWatcher adds/removes managers from it as kubeconfig secrets are added and removed.

I chose this approach because it hides internals from the user by preventing them from manually adding and removing managers to the set without hiding that the manager set is the canonical source of per-cluster managers. That said, I don't love the setup here, and would appreciate any additional feedback.

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,109 @@
...
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const (
+	LocalCluster = ""
+)
+
+// ClusterHandler is passed to RunClusterWatcher to handle select cluster events.
+type ClusterHandler interface {
```
ilackarms, line 22: explain this interface is implemented internally and exposed for custom use cases

pkg/multicluster/watcher.go
```diff
@@ -18,33 +19,24 @@ const (
...
-	managers ManagerSet
+	cancels  *cancelSet
 }
 
 // RunClusterWatcher initializes and runs a reconciler for KubeConfig secrets.
-// It starts and runs provided ClusterHandlers for the localManager, and maintains a set of
-// active cluster manager.Managers in mangers.
-func RunClusterWatcher(ctx context.Context, localManager manager.Manager, managers ManagerSet, handlers ...ClusterHandler) error {
+// It starts and runs ClusterHandlers for the localManager as if it were discovered by the watcher.
+func RunClusterWatcher(ctx context.Context, localManager manager.Manager, handlers ...ClusterHandler) error {
```
ilackarms, line 30: nit: what we are referring to as `localManager` might not in fact be a local manager. this `mgr` simply belongs to the cluster where we expect secrets. maybe masterManager?

ilackarms, line 30: nit: dyt `WatchClusterSecrets` is a more descriptive name?

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,140 @@
...
+
+	return reconcile.Result{}, c.startManager(clusterName, mgr)
+}
+
+func (c *clusterWatcher) startManager(clusterName string, mgr manager.Manager) error {
+	ctx, cancel := context.WithCancel(
+		contextutils.WithLoggerValues(context.WithValue(c.ctx, "cluster", clusterName), zap.String("cluster", clusterName)))
+	go func() {
+		err := mgr.Start(ctx.Done())
+		if err != nil {
```
ilackarms, line 73: `err` will be non-nil if `ctx` is cancelled, causing an unwanted panic here

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,140 @@
...
+			cancel()
+			contextutils.LoggerFrom(ctx).DPanicw("manager start failed for cluster %v", clusterName)
+		}
+	}()
+
+	c.cancels.set(clusterName, cancel)
+
+	errs := &multierror.Error{}
+	for _, handler := range c.handlers {
+		err := handler.HandleAddCluster(ctx, clusterName, mgr)
```
ilackarms, line 83: does this function need to return an error?

pkg/multicluster/handler.go
```diff
@@ -0,0 +1,16 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// ClusterHandler is passed to RunClusterWatcher to handle select cluster events.
+// It is implemented internally by skv2 components but can be implemented by the user for specialized use cases.
+type ClusterHandler interface {
+	// ClusterHandler is called when a new cluster is identified by a cluster watch.
+	// The provided context is cancelled when a cluster is removed, so any teardown behavior for removed clusters
+	// should take place when ctx is cancelled.
+	HandleAddCluster(ctx context.Context, cluster string, mgr manager.Manager) error
```
ilackarms, line 15: does this method need to return an error?

also nit, `HandleAddCluster` > `AddCluster`

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,140 @@
...
+		HealthProbeBindAddress: "0",
+	})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, c.startManager(clusterName, mgr)
+}
+
+func (c *clusterWatcher) startManager(clusterName string, mgr manager.Manager) error {
```
ilackarms, line 68: does this function need to return an error?

pkg/multicluster/watcher.go
EItanya: This isn't true, if a stop signal is received, the manager will return nil https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/manager/internal.go#L430

pkg/multicluster/handler.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// ClusterHandler is passed to RunClusterWatcher to handle select cluster events.
+// It is implemented internally by skv2 components but can be implemented by the user for specialized use cases.
+type ClusterHandler interface {
```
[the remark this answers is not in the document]
joekelley: To this point i've been able to address all use cases by waiting on the context to be cancelled, but i'm not opposed to adding a remove func if it becomes necessary

pkg/multicluster/reconcile_loop.go
```diff
@@ -0,0 +1,94 @@
...
+
+type DeletionReconciler interface {
+	// we received a reconcile request for an object that was removed from the cache
+	ReconcileDeletion(cluster string, request reconcile.Request)
+}
+
+// a Reconcile Loop runs resource reconcilers until the context gets cancelled
+type Loop interface {
+	// RunReconciler adds a reconciler to a slice of reconcilers that will be run against
+	RunReconciler(ctx context.Context, reconciler Reconciler, predicates ...predicate.Predicate) error
```
joekelley, line 28: I think I'll rename this since it doesn't really "Run" anything in this case, just add to a slice. Something like "RegisterReconciler", open to suggestions.

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+)
+
+type ClusterWatcher interface {
+	// Run starts a watch for KubeConfig secrets on the cluster managed by the local manager.Manager.
+	// Note that Run will call Start on the local manager and run all registered ClusterHandlers.
+	Run(local manager.Manager) error
+	// RegisterClusterHandler adds a ClusterHandler to the ClusterWatcher.
+	RegisterClusterHandler(handler ClusterHandler)
+	// Stop cancels the context on every manager.Manager.
+	Stop()
```
ilackarms, line 28: can we just cancel the context passed to the clusterwatcher?

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+		return reconcile.Result{}, eris.Wrap(err, "failed to extract kubeconfig from secret")
+	}
+
+	// If the cluster already has a manager, remove the existing instance and start again.
+	if _, err := c.cancels.get(clusterName); err == nil {
+		c.removeCluster(clusterName)
+	}
+
+	mgr, err := manager.New(cfg.RestConfig, manager.Options{
+		// TODO these should be configurable, disable for now
```
ilackarms, line 67: my suggestion here is to allow the user to provider a `manager.Options` in the cw constructor. if none provided, use a default

EItanya: My one push back is the metrics bind addresses. The unification of metrics still needs to be tackled of course, but for the time being I think we should force these values as zero here, even if we accept options for everything else.

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
```diff
@@ -0,0 +1,89 @@
+// Definitions for the multicluster Kubernetes Controllers
+package multicluster
```
EItanya, line 2: Can we add some suffix related to the group, so each one doesn't have the same package name. Possibly the version?

test/test_config.go
```diff
@@ -38,3 +43,9 @@ func ManagerWithOpts(cfg *rest.Config, opts manager.Options) (manager.Manager, f
 
 	return mgr, cancel
 }
+
+func ManagerNotStarted(opts manager.Options) manager.Manager {
```
EItanya, line 47:
```suggestion
func MustManagerNotStarted(opts manager.Options) manager.Manager {
```

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+	Run(local manager.Manager) error
+	// RegisterClusterHandler adds a ClusterHandler to the ClusterWatcher.
+	RegisterClusterHandler(handler ClusterHandler)
+	// Stop cancels the context on every manager.Manager.
+	Stop()
+}
+
+type clusterWatcher struct {
+	ctx      context.Context
+	mutex    sync.RWMutex
```
EItanya, line 33: unused

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+	return reconcile.Result{}, nil
+}
+
+func (c *clusterWatcher) startManager(clusterName string, mgr manager.Manager) {
+	ctx, cancel := context.WithCancel(
+		contextutils.WithLoggerValues(context.WithValue(c.ctx, "cluster", clusterName), zap.String("cluster", clusterName)))
+	go func() {
+		err := mgr.Start(ctx.Done())
+		if err != nil {
+			contextutils.LoggerFrom(ctx).DPanicw("manager start failed for cluster %v", clusterName)
```
EItanya, line 86: Do we want to somehow propagate this error? We are using an `ErrorSignaler` for this in SMH
https://github.com/solo-io/service-mesh-hub/tree/master/pkg/errutils

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+func (c *clusterWatcher) ReconcileSecret(obj *v1.Secret) (reconcile.Result, error) {
+	clusterName, cfg, err := kubeconfig.SecretToConfig(obj)
+	if err != nil {
+		return reconcile.Result{}, eris.Wrap(err, "failed to extract kubeconfig from secret")
+	}
+
+	// If the cluster already has a manager, remove the existing instance and start again.
+	if _, err := c.cancels.get(clusterName); err == nil {
```
EItanya, line 62: Do we want to verify that something meaningful about the secret has changed and not just metadata?

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,193 @@
...
+		contextutils.LoggerFrom(c.ctx).Debugw("reconciled delete on cluster secret for uninitialized cluster %v", clusterName)
+		return
+	}
+
+	cancel()
+	c.cancels.delete(clusterName)
+}
+
+func (c *clusterWatcher) RegisterClusterHandler(handler ClusterHandler) {
+	c.handlers.add(handler)
```
EItanya, line 112: Do we want to retroactively add existing clusters?

pkg/multicluster/kubeconfig/predicate.go
EItanya: Instead you could use `predicate.Funcs` and just return an instance, rather than having a public struct

pkg/multicluster/watcher.go
joekelley: good idea, this can be handled in the predicate

pkg/multicluster/kubeconfig/predicate.go
```diff
@@ -1,31 +1,34 @@
...
+			return isKubeConfigSecret(e.Object)
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return isKubeConfigSecret(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if isKubeConfigSecret(e.ObjectNew) {
+				// Ignore metadata changes.
+				oldSecret, newSecret := e.ObjectOld.(*corev1.Secret), e.ObjectNew.(*corev1.Secret)
+				return !reflect.DeepEqual(oldSecret.Data, newSecret.Data)
```
EItanya, line 24: If we are worried about using reflect here, we can use a combination of checking for key equality, as well as `bytes.Equals`

joekelley: my understanding of DeepEqual was that it was performant for like and simple types, but I'm open to handwriting the equality function if that assumption is wrong

pkg/multicluster/watcher.go
joekelley: Reading this as "call the new ClusterHandler's AddCluster function for all clusters already discovered by the watcher", disregard if I misunderstood.

good question, any thoughts on this @ilackarms ? Currently the watcher is not responsible for maintaining the active set of managers, just their cancel functions, so we'd need a refactor to support this behavior.

A multicluster client will be introduced in a follow-up PR. The plan there is to have the client register with the watch as a ClusterHandler and maintain the set of managers created by the ClusterWatcher. If we decide we want the feature proposed here, it may make sense for the watcher to be responsible for storing active managers and adding a function to expose them.

pkg/multicluster/watcher.go
joekelley: Can you point me to where the propagated errors are used? I like the idea but am reluctant to include an error signaler in this PR as it's already pretty large and I can't think of a way to programatically handle these errors beyond panicking and logging.

pkg/multicluster/kubeconfig/predicate.go
EItanya: No should be fine here

pkg/multicluster/watcher.go
EItanya: In the [`AsyncManager`](https://github.com/solo-io/service-mesh-hub/blob/master/services/common/multicluster/manager/manager.go#L121). If an error occurs while starting the manager, it sends it to the signaler, which the caller can listen to.
Like so: https://github.com/solo-io/service-mesh-hub/blob/master/services/common/multicluster/common.go#L58

pkg/multicluster/watcher.go
EItanya: So I think it depends whether we think handlers should have all previously added clusters, added upon startup. If so I would definitely lean towards having it here. If we push it down a level, then that level becomes the new root, and it sounds like we want the multicluster client to be just like any other handler.

To solve this, we could just change the cancelSet, to a set containing a struct such as:
```
type ContextualManager struct {
    ctx context.Context
    cancel context.CancelFunc
    mgr manager.manager
}
```

This would also allow flexibility to add things like the errSignaler I mentioned above.

pkg/multicluster/watcher.go
joekelley: Right, but we ultimately fatal if an error occurs: https://github.com/solo-io/service-mesh-hub/blob/2278e78e628302ac3e7fe5858254da4fc31f4848/services/mesh-discovery/run.go#L55.

I'd prefer to keep things simple for now and add a signaler later if there's more that we can do with the error received.

pkg/multicluster/watcher.go
joekelley: updated to allow passing in a manager.Options. metrics and health check addresses are still zeroed out by a `default` function before creating the managers. let me know if that's not how we want to handle this.

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
joekelley: On second thought i'm a little reluctant to do this because then we have directory structures like `group.solo.io/v1/controller_v1/multicluster_v1` which reads redundant IMO, and since that doesn't guarantee uniqueness when there are two groups with a v1 I don't think it's buying us much over leaving the user to alias.

pkg/multicluster/watcher.go
joekelley: I made these changes. It necessitated the addition of a ClusterHandler per reconciler.

Now, when adding a reconcile loop, we first register the loop itself with the ClusterWatcher. The loop is a ClusterHandler whose AddCluster function creates a reconcile.Loop for every cluster. Its AddReconciler function now creates a new object called a `reconcilerRunner` which also implements the ClusterHandler interface, and registers it to the ClusterWatcher. This allows us to add reconcilers to a multicluster reconcile loop after the ClusterWatcher has been started.

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,216 @@
...
+		return reconcile.Result{}, err
+	}
+
+	c.startManager(clusterName, mgr)
+
+	return reconcile.Result{}, nil
+}
+
+func (c *clusterWatcher) ReconcileSecretDeletion(req reconcile.Request) {
+	// TODO update to namespace.name
```
joekelley, line 88: This vague comment alludes to updating our concept of "cluster name" from the name on the kubeconfig secret to the namespace.name on the kubeconfig secret.

While this would eliminate any risk of name duplication, it's also a bit cumbersome, and leaks the detail of how clusters are stored into all places we reference clusters. This includes code and any resource with a multicluster ref. I want to be clear on how we intend to refer to clusters in the long term before committing to this approach.

pkg/multicluster/watcher.go
```diff
@@ -0,0 +1,216 @@
...
+	Run(master manager.Manager) error
+	// RegisterClusterHandler adds a ClusterHandler to the ClusterWatcher.
+	RegisterClusterHandler(handler ClusterHandler)
+}
+
+// ManagerSet maintains a manager for every cluster in the system.
+type ManagerSet interface {
+	// Cluster returns a manager for the given cluster, or an error if one does not exist.
+	Cluster(cluster string) (manager.Manager, error)
+}
```
EItanya, line 36: Do we think this is necessary, in SMH we have never needed the whole manager for a different config, just the client. Maybe we don't need to implement this interface.

joekelley: I was thinking that the multicluster client would depend on ManagerSet. something like:

```go
type Client struct {
  managers ManagerSet
}

func (c Client) Cluster(cluster string) (client.Client, error) {
  return c.managers.Cluster(string).GetClient()
}
```

Maybe we don't need that intermediary though, and the ClusterWatcher itself could implement multicluster.Client without exposing the managers themselves.

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
ilackarms: @joekelley why not just leave it in `controller` package?

pkg/multicluster/kubeconfig/predicate.go
```diff
@@ -0,0 +1,39 @@
...
+import (
+	"reflect"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+func Predicate() predicate.Predicate {
```
ilackarms, line 12: how about using a `var Predicate = ` here instead of a `func`?

ilackarms: as a convention for storing global types (e.g. it allows modifyting the `Predicate` for testing)

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
joekelley: this was done to deal with an import cycle, i'll see if there's a way to bring it back into the controller pkg

joekelley:
```
package github.com/solo-io/skv2/codegen/render_test
	imports github.com/solo-io/skv2/codegen/test/api/things.test.io/v1/controller
	imports github.com/solo-io/skv2/pkg/multicluster
	imports github.com/solo-io/skv2/pkg/generated/kubernetes/core/v1/controller
	imports github.com/solo-io/skv2/pkg/multicluster
```

pkg/multicluster/watcher.go
joekelley: cleaned up https://github.com/solo-io/skv2/pull/4/commits/b6f7f2b2bccc31f7ce22c03715f596f008844f81

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
joekelley: gonna fix the cycle

pkg/multicluster/watcher.go
joekelley: service mesh hub overrides the same options so i'll stick with this.

pkg/multicluster/watcher.go
joekelley: will keep this as is per Eitan's comment

codegen/templates/code/controller/multicluster_reconcilers.gotmpl
joekelley: will resolve the cycle before merging

pkg/multicluster/watcher.go
joekelley: going to remove the todo from code and file an issue about fleshing out our cluster registry story

pkg/multicluster/watcher.go
joekelley: I like the added flexibility of exposing the manager here in case use cases arise. The multicluster client will expose just the client in order to simplify consumption for the common use case.

pkg/multicluster/reconcile/reconcile.go
```diff
@@ -39,14 +39,14 @@ func NewLoop(name string, cw multicluster.ClusterWatcher, resource ezkube.Object
 // AddCluster creates a reconcile loop for the cluster.
 func (r *clusterLoopRunner) AddCluster(ctx context.Context, cluster string, mgr manager.Manager) {
 	loopForCluster := reconcile.NewLoop(r.name+"-"+cluster, mgr, r.resource)
+
+	// Cleanup any potentially stale references to the cluster.
+	r.clusterLoops.remove(cluster)
+	r.reconcilers.unset(cluster)
```
joekelley, line 45: If we clean up the cluster loop and reconcile sets like this, it's possible for the cluster loop map in r.clusterLoops and cluster sets in r.reconcilers to grow indefintiely as clusters are removed from the system.

Fixes for this include the goroutine approach this replaced or adding an optional RemoveCluster function to the cluster handler.

joekelley: https://github.com/solo-io/skv2/pull/4/commits/17cf19cc0042d6d49e9e2a971410ccbea1550f4c

pkg/multicluster/handler.go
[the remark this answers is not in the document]
joekelley: found a use case : ) https://github.com/solo-io/skv2/pull/4/commits/17cf19cc0042d6d49e9e2a971410ccbea1550f4c

pkg/multicluster/reconcile.go
```diff
@@ -0,0 +1,26 @@
...
+type Reconciler interface {
+	// reconcile an object
+	// requeue the object if returning an error, or a non-zero "requeue-after" duration
+	Reconcile(cluster string, object ezkube.Object) (reconcile.Result, error)
+}
+
+type DeletionReconciler interface {
+	// we received a reconcile request for an object that was removed from the cache
+	ReconcileDeletion(cluster string, request reconcile.Request)
+}
```
EItanya, line 20: Just wondering out loud why the decision was made to split this interface, and allow the `DeletionReconciler` to be optional. In general I find optional interfaces to be hard to reason about. Instead could we combine them and leave it up to the user whether they want to implement the delete in their code, or just leave it empty.

EItanya: Discussed this offline, reasoning makes sense

joekelley: discussed offline

PR #20, opened 2020-04-27
Cluster register

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/21


pkg/multicluster/auth/interfaces.go
```diff
@@ -0,0 +1,66 @@
...
+
+	k8s_core_v1 "github.com/solo-io/skv2/pkg/generated/kubernetes/core/v1"
+	k8s_core_types "k8s.io/api/core/v1"
+	k8s_rbac_types "k8s.io/api/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+//go:generate mockgen -source ./interfaces.go -destination ./mocks/mock_auth.go
+
+type RbacBinder interface {
```
GrahamGoudeau, line 14: Hard -1 to letting this live in two separate places. There's ongoing churn in this section of code and I do not want to maintain it in two separate repos.

I'm not going to approve this until we have a clear ownership/path to integrating it back into SMH

GrahamGoudeau: To be clear: By "this" I mean the entirety of this PR, not this file in particular

pkg/multicluster/auth/config_creator.go
```diff
@@ -0,0 +1,104 @@
...
+		if err != nil {
+			return err
+		}
+
+		if len(serviceAccount.Secrets) == 0 {
+			return eris.Errorf("service account %s.%s does not have a token secret associated with it", name, namespace)
+		}
+		if len(serviceAccount.Secrets) != 1 {
+			return eris.Errorf("service account %s.%s unexpectedly has more than one secret", name, namespace)
+		}
```
joekelley, line 90: would this qualify as a SecretNotReady error, or would the user have to do something to remove the extra secret in order to continue?

pkg/multicluster/kubeconfig/kubeconfig_secret_test.go
```diff
@@ -46,6 +46,7 @@ users:
 	})
 
 	Describe("ToSecret", func() {
```
joekelley, line 48: nit: let's add new describe blocks for the new functions "ToSecretWithKey" and "SecretToConfigWIthKey"

pkg/generated/generate.go
```diff
@@ -21,6 +23,14 @@ var (
 
 func main() {
 	log.Println("starting kube client generation")
+
+	// load custom client providers template
+	customClientProvidersBytes, err := ioutil.ReadFile("templates/custom_providers.gotmpl")
```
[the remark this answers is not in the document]
EItanya: So the reason I need that template is for the `FromConfigFactory`, not for the `FromConfig` itself. The ClusterAuthorizer needs to build 3 different clients for the remote cluster. Originally I thought of just building one `client.Client` and then using that to construct the other clients, but that isn't mockable.

pkg/multicluster/auth/cluster_authorization.go
```diff
@@ -0,0 +1,196 @@
...
+) (bearerToken string, err error) {
+
+	saToCreate := &core_v1.ServiceAccount{
+		ObjectMeta: k8s_meta.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+
+	if err = c.serviceAccountClient.UpsertServiceAccount(ctx, saToCreate); err != nil {
```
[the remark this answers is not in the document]
EItanya: Good point, What do you think the behavior should be?

pkg/multicluster/auth/config_creator.go
```diff
@@ -0,0 +1,104 @@
...
+
+		secretName := serviceAccount.Secrets[0].Name
+
+		secret, err := r.secretClient.GetSecret(ctx, client.ObjectKey{Name: secretName, Namespace: namespace})
+		if err != nil {
+			return err
+		}
+
+		foundSecret = secret
+		return nil
```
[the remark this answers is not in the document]
EItanya: cleaned up a bit

pkg/multicluster/auth/config_creator.go
EItanya: According to the kube API, multiple secrets can exist, so it seems safe to just take the first one https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#serviceaccount-v1-core

ci/check-code-and-docs-gen.sh
```diff
@@ -16,8 +16,6 @@ if [ ! -f .gitignore ]; then
   echo "_output" > .gitignore
 fi
 
-make install-deps
```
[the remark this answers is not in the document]
EItanya: no because it is already called by the `generated-code` target https://github.com/solo-io/skv2/blob/ad7e6dee5b8f735967358f0004d57a71d9b39ba8/Makefile#L24

pkg/multicluster/auth/cluster_authorization.go
```diff
@@ -0,0 +1,202 @@
...
+	core_v1 "k8s.io/api/core/v1"
+	k8s_rbac_types "k8s.io/api/rbac/v1"
+	k8s_meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+var (
+	// visible for testing
+	ServiceAccountClusterAdminRoles = []*k8s_rbac_types.ClusterRole{{
```
joekelley, line 19: nit: move to test package

api/generate.go
```diff
@@ -47,6 +48,7 @@ func main() {
 				RenderClients:    true,
 				RenderTypes:      true,
 				ApiRoot:          "pkg/api",
+				CustomTemplates:  contrib.AllCustomTemplates,
```
ilackarms, line 51: i'm not sure we want `AllCustomTemplates` here, since we'll wind up with unused code as things get added to `contrib`

contrib/custom_templates.go
```diff
@@ -19,16 +19,23 @@ const (
...
 	setsTemplates := model.CustomTemplates{
 		Templates: map[string]string{SetOutputFilename: templateContents},
 	}
-	// register sets
-	AllCustomTemplates = append(AllCustomTemplates, setsTemplates)
 
-	return setsTemplates
-}()
+	return setsTemplates, nil
+}
```
ilackarms, line 41: why this change? honestly what i like about the other model is how easily copy-pasteable it is. this will require editing functions in 2 places

EItanya: The other way was a pattern I had never seen before. It was attempting to use `init()` functionality without actually using an `init()` function. I personally think this version is more immediately understandable, and closer to common go paradigms.

pkg/generated/templates/custom_providers.gotmpl
```diff
@@ -0,0 +1,37 @@
...
+func {{ $resource.Kind }}ClientFromConfigFactoryProvider() {{ $resource.Kind }}ClientFromConfigFactory {
+return func(cfg *rest.Config) ({{ $resource.Kind }}Client, error) {
+clients, err := NewClientsetFromConfig(cfg)
+if err != nil {
+return nil, err
+}
+return clients.{{ pluralize $resource.Kind }}(), nil
+}
+}
+{{- end }}
```
ilackarms, line 37: can the indentation on this file be fixed?

pkg/generated/parallel_mockgen.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+# K8s config
+mockgen -package mock_clientcmd -destination ./mocks/k8s/clientcmd/config.go k8s.io/client-go/tools/clientcmd ClientConfig &
+
+# K8s clients
+mockgen -package mock_k8s_core_clients -destination ./kubernetes/mocks/core/v1/clients.go github.com/solo-io/skv2/pkg/generated/kubernetes/core/v1 Clientset,ServiceClient,PodClient,NamespaceClient,NodeClient,ServiceAccountClient,SecretClient,ConfigMapClient &
+mockgen -package mock_k8s_rbac_clients -destination ./kubernetes/mocks/rbac.authorization.k8s.io/v1/clients.go github.com/solo-io/skv2/pkg/generated/kubernetes/rbac.authorization.k8s.io/v1 Clientset,ClusterRoleBindingClient,RoleBindingClient,ClusterRoleClient,RoleClient &
+
+wait
```
ilackarms, line 10: wondering if we should just generate our clientsets with mockgen `go:generate` targets instead of this

pkg/generated/templates/custom_providers.gotmpl
```diff
@@ -0,0 +1,37 @@
...
+import (
+"context"
+
+"k8s.io/apimachinery/pkg/api/errors"
+"k8s.io/client-go/kubernetes/scheme"
+"k8s.io/client-go/rest"
+"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+{{- range $resource := $.Resources }}
```
ilackarms, line 13: please add a comment here explaining the intention / brief usage of this file

ilackarms: (above the `{{- range $resource := $.Resources }}`)

pkg/multicluster/kubeconfig/kubeconfig_secret.go
```diff
@@ -31,22 +52,44 @@ func ToSecret(namespace string, cluster string, kc api.Config) (*kubev1.Secret,
...
-// Config contains various KubeConfig formats for convenience.
-type Config struct {
-	ClientConfig clientcmd.ClientConfig
-	ApiConfig    *api.Config
-	RestConfig   *rest.Config
+// https://github.com/solo-io/service-mesh-hub/issues/590
+// If the user has a cert authority file set instead of the raw bytes in their kubeconfig, then
+// we'll fail later when the pods in-cluster try to read that file path.
+// We need to read the file right now, in a CLI context, and manually shuffle the bytes over to the CA data field
+func readCertAuthFileIfNecessary(cfg api.Config) error {
```
ilackarms, line 63: i'd add an explanation here of why this is necessary (**we are reading in a kubeconfig stored on a user's local machine, which may not use inline certs**)

i'd maybe rename the function to `convertCertFilesToInline`

pkg/multicluster/register/internal/cluster_rbac_binder.go
```diff
@@ -0,0 +1,143 @@
...
+	k8s_meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	EmptyRolesListError        = eris.New("Empty Roles list found, must specify at least one role to bind to.")
+	EmptyClusterRolesListError = eris.New("Empty ClusterRoles list found, must specify at least one role to bind to.")
+)
+
```
ilackarms, line 20: it's still not clear to me what the code in this file is used for. it would be helpful to comment what the `clusterAuthorization` does and where it is called / used

EItanya: `clusterAuthorization` is a leftover from a previous iteration. This should be named `clusterRbacBinder`. It is the implementation of the `ClusterRBACBinder` interface.

PR #22, opened 2020-04-27
multicluster kind e2e

joekelley:


codegen/render/kube_multicluster_test.go
```diff
@@ -0,0 +1,403 @@
...
+		logLevel      = zap.NewAtomicLevel()
+		ctx           = context.TODO()
+		remoteContext = os.Getenv("REMOTE_CLUSTER_CONTEXT")
+	)
+
+	BeforeEach(func() {
+		logLevel.SetLevel(zap.DebugLevel)
+		log.SetLogger(zaputil.New(
+			zaputil.Level(&logLevel),
+		))
```
EItanya, line 54: Is this just to silence the logger?

codegen/render/kube_multicluster_test.go
```diff
@@ -0,0 +1,403 @@
...
+	m     map[string]*Paint
+	mutex sync.Mutex
+}
+
+func newConcurrentPaintMap() *concurrentPaintMap {
+	return &concurrentPaintMap{
+		m:     make(map[string]*Paint),
+		mutex: sync.Mutex{},
+	}
+}
```
EItanya, line 365: Since this is test code anyway, maybe we should use https://golang.org/src/sync/map.go rather than creating all this boilerplate

.github/workflows/pull_request.yaml
```diff
@@ -63,4 +68,4 @@ jobs:
           make install-deps
       - name: Run tests
         run: |
-          go test ./...
\ No newline at end of file
+          REMOTE_CLUSTER_CONTEXT=kind-skv2-test-remote go test ./...
```
EItanya, line 71: Should we maybe run the kind script from within the test somehow so that the context name isn't a magic string from inside of the shell script

codegen/render/kube_multicluster_test.go
joekelley: it sets logs to debug. this is a copy/paste from our other kube tests in this package, but it helped a lot with debugging the multicluster watcher becuase i could see which controllers were being started and stopped over the course of a test

.github/workflows/pull_request.yaml
joekelley: Going to make the remote cluster context value configurable such that this string is passed to both the setup script and this environment variable. This allows us to be less prescriptive about how users set up their local test environments / manage their kube contexts while relaxing the magic here.

codegen/render/kube_multicluster_test.go
joekelley: good tip! ended up using this to simplify the boilerplatey concurrent maps implemented here, as their API is a bit more readable in the context of the tests imo

Makefile
```diff
@@ -16,7 +16,7 @@ install-deps: mod-download
 	go get -v github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc
 	go get -v github.com/solo-io/protoc-gen-ext
 	go get -v github.com/golang/mock/mockgen
-	go get -v golang.org/x/tools/cmd/goimports
+	go get -v golang.org/x/tools/cmd/goimports@v0.0.0-20200427205912-352a5409fae0
```
EItanya, line 19: THANK! GOD! 🙏

PR #24, opened 2020-04-28
Crd schema

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/145


EItanya:
Update on the state of this PR:

This PR is meant to parse our proto based kube APIs into OpenApiV3 schemas for the purpose of CRD schema validation. To do this it attempts to use [`cue`](https://github.com/cuelang/cue) as a library to generate the schema. This is the same method istio uses to generate their schemas. https://github.com/istio/tools/tree/master/cmd/cue-gen. Originally I was hoping to use the istio cli directly, but there are too many assumptions built in and it will not work for us. 

In order to use cue as a library also requires minor changes to the `solo-kit` `collector`. The `collector` is the component which figures out all of the imports for a given proto file, and compiles it using the `protoc` compiler. In order to remove any dependency on `solo-kit` this PR moves that code into `skv2` and slightly changes the interface to allow for discovery of all imports for a given proto file, without calling the collector as a whole. All of those changes can be found in `codegen/collector`. 

`codegen/cmd.go` has a very rough example of what implementing this might look like. Currently it just generates the Schema and does nothing with it. In the future this schema will have to be added to the `CRD` manifest.

ci/tools.go
```diff
@@ -17,7 +17,6 @@ limitations under the License.
 package tools
 
 import (
-	_ "github.com/envoyproxy/protoc-gen-validate"
```
EItanya, line 20: Any specific reason for this?

go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/solo-io/solo-kit v0.14.0
```
EItanya, line 36: Can we now get rid of the `solo-kit` dependency altogether?

codegen/skv2_anyvendor/defaults.go
```diff
@@ -0,0 +1,70 @@
...
+	ExtProtoMatcher = &anyvendor.GoModImport{
+		Package:  "github.com/solo-io/protoc-gen-ext",
+		Patterns: []string{"extproto/*.proto"},
+	}
+
+	// matches all solo-kit protos, useful for any projects using solo-kit
+	// SoloKitProtoMatcher = &anyvendor.GoModImport{
+	// 	Package:  "github.com/solo-io/solo-kit",
+	// 	Patterns: []string{"api/**/*.proto"},
+	// }
```
EItanya, line 21: Delete comments

codegen/skv2_anyvendor/defaults.go
```diff
@@ -0,0 +1,70 @@
...
+	// matches gogo.proto, used for gogoproto code gen.
+	GogoProtoMatcher = &anyvendor.GoModImport{
+		Package:  "github.com/gogo/protobuf",
+		Patterns: []string{"gogoproto/*.proto", "protobuf/**/*.proto"},
+	}
+
+	// default match options which should be used when creating a solo-kit project
+	DefaultExternalMatchOptions = map[string][]string{
+		ExtProtoMatcher.Package: ExtProtoMatcher.Patterns,
+		// SoloKitProtoMatcher.Package: SoloKitProtoMatcher.Patterns,
```
EItanya, line 32: delete

codegen/model/resource.go
```diff
@@ -56,6 +56,9 @@ type Group struct {
 	// Should we generate kubernetes manifests?
 	RenderManifests bool
 
+	// Should we generate validating schemas for CRDs?
```
EItanya, line 59: Do you think this should be true or false by default? I can't decide

harveyxia: Given that there may still be protobuf constraints with the cue parser, I think it should be false by default so that it doesn't break code generation for downstream projects that upgrade skv2.

ci/tools.go
harveyxia: hmm i don't know why this got removed, i'll restore it

codegen/render/manifests_renderer.go
```diff
@@ -3,27 +3,36 @@ package render
...
-
+	"github.com/solo-io/skv2/codegen/model"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 )
 
 // creates a k8s resource for a group
 // this gets turned into a k8s manifest file
-type MakeResourceFunc func(group Group) []metav1.Object
+type MakeResourceFunc func(group Group, oapiSchemas kuberesource.OpenApiSchemas) ([]metav1.Object, error)
```
EItanya, line 21: Rather than change this function, should we put the schemas on the `Group`?

codegen/kuberesource/crd.go
```diff
@@ -1,26 +1,108 @@
...
+	}
+	statusSchema, err := getJsonSchema(resource, resource.Status.Type.Name, oapiSchemas)
+	if err != nil {
+		return nil, eris.Wrapf(err, "constructing status validation schema for Kind %s", resource.Kind)
+	}
+	return &apiextv1beta1.CustomResourceValidation{
+		OpenAPIV3Schema: &apiextv1beta1.JSONSchemaProps{
+			Type: "object",
+			Properties: map[string]apiextv1beta1.JSONSchemaProps{
+				"spec":   *specSchema,
```
EItanya, line 51: This works because of the expanded references?

codegen/model/resource.go
EItanya: Makes sense to me

codegen/render/manifests_renderer.go
harveyxia: I'll defer to @ilackarms for how best to represent this

codegen/kuberesource/crd.go
harveyxia: Yup, exactly, the generated openapi schemas are fully self contained because `ExpandReferences: true` on the cuelang cfg

codegen/kuberesource/crd.go
```diff
@@ -1,26 +1,108 @@
...
+	}
+	return objects, nil
+}
+
+func constructValidationSchema(resource model.Resource, oapiSchemas OpenApiSchemas) (*apiextv1beta1.CustomResourceValidation, error) {
+	specSchema, err := getJsonSchema(resource, resource.Spec.Type.Name, oapiSchemas)
+	if err != nil {
+		return nil, eris.Wrapf(err, "constructing spec validation schema for Kind %s", resource.Kind)
+	}
+	statusSchema, err := getJsonSchema(resource, resource.Status.Type.Name, oapiSchemas)
```
ilackarms, line 43: need to ensure `resource.Status` is not nil here

codegen/model/resource.go
```diff
@@ -56,6 +56,11 @@ type Group struct {
...
+	// NOTE: proto int64's cannot currently be used with generated validation schemas.
+	// See https://github.com/solo-io/skv2/issues/146 for details.
+	RenderValidationSchemas bool
```
ilackarms, line 62: it would be nice to add a test for this (in cmd_test.go), and would also be nice to enable it for the skv2 multicluster api group

codegen/render/manifests_renderer.go
ilackarms: i tend to agree with Eitan, i think it's better to move `oapiSchemas` into the place we care about them (inside the CRD constructor function). though it seems we will need to put `protoDir` here

ilackarms: would that cause us to run the `collector` an additional time for every group?

codegen/render/manifests_renderer.go
```diff
@@ -37,9 +46,19 @@ func (r ManifestsRenderer) RenderManifests(grp Group) ([]OutFile, error) {
 	if !grp.RenderManifests {
 		return nil, nil
 	}
+
+	var err error
```
ilackarms, line 50: just a nit, but in cases like this i like to ut the `var err` declaration inside the block where it's read

codegen/skv2_anyvendor/defaults.go
```diff
@@ -0,0 +1,63 @@
...
+			},
+		})
+	}
+	return &anyvendor.Config{
+		Local: &anyvendor.Local{
+			Patterns: i.Local,
+		},
+		Imports: imports,
+	}
+}
```
ilackarms, line 63: did we not already have this file here?

codegen/render/manifests_renderer.go
harveyxia: Moved to a field in `Group`. And no, this would not cause the collector run an additional time per group.

pkg/jsonpb/README.md
```diff
@@ -0,0 +1,15 @@
+This package is copied from github.com/gogo/protobuf at commit deb6fe8ca7c6d06584bfbd40ca407bf69d9fd2aa.
+
+
+## Customizations
+
+1. Serialize `int64` and `uint64` to integers, not strings.
```
EItanya, line 6: Is all of this to be able to use `int64` and `uint64`? I think I would argue we just disallow those types rather than attempt to maintain our own marshallers and unmarshallers. Just wondering how we decided to go down this path

ilackarms: as per discussing with @yuval-k we have certain int64's in gloo that we can't get rid of. discussed and determined that the cost of maintaining a fork of `jsonpb` would be minimal

codegen/render/kube_crud_test.go
```diff
@@ -69,7 +69,7 @@ func mustCrud(ctx context.Context, clientSet Clientset, paint *Paint) {
 	Expect(written.Spec).To(Equal(paint.Spec))
 
 	status := PaintStatus{
-		ObservedGeneration: written.Generation,
+		ObservedGeneration: uint32(written.Generation),
```
ilackarms, line 72: is this correct? no longer needed, correct?

harveyxia: correct, i restored the int64's

ilackarms (approved):

PR #31, opened 2020-05-04
support basic name/value env vars in generated helm charts

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/30

### DIFF
--- a/codegen/templates/chart/values.yamltmpl
+++ b/codegen/templates/chart/values.yamltmpl
@@ -11,4 +11,8 @@
     [[ $port.Name ]]: [[ $port.DefaultPort ]]
 [[- end ]]
 [[- end ]]
+[[- if gt (len $operator.Env) 0 ]]
+  env:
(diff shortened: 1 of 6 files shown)


codegen/templates/chart/values.yamltmpl
```diff
@@ -11,4 +11,8 @@
     [[ $port.Name ]]: [[ $port.DefaultPort ]]
 [[- end ]]
 [[- end ]]
+[[- if gt (len $operator.Env) 0 ]]
+  env:
+[[ toYaml $operator.Env | indent 4 ]]
+[[- end ]]
```
joekelley, line 17: With this approach the env section of the values file looks just like the template, which could lead to some redundancy, but I like that it matches the kube EnvVar API and it will be familiar to users. Plus this way users can opt to get a value from a configmap/secret/etc rather than our prescribed location.

PR #32, opened 2020-05-05
fix e2e test helper to set context on client config

joekelley:
Fix e2e test helper such that context is set on resulting client config. Also update multicluster reconciler now that this bug is fixed, and the tests actually run against two clusters.


joekelley:
here's the meat of the change: https://github.com/solo-io/skv2/pull/32/files#diff-02c7c5f7bde2aab57cce1e65c9e2c56bR67

PR #37, opened 2020-05-11
Common api

EItanya:


api/core/v1alpha1/core.proto
```diff
@@ -0,0 +1,42 @@
+syntax = "proto3";
+
+package core.skv2.solo.io;
+
+option go_package = "github.com/solo-io/skv2/pkg/api/core.skv2.solo.io/v1alpha1/types";
```
ilackarms, line 5: not sure i agree with the choice of v1alpha1 for the version for core types. 

also, what's the reasoning using a `types` package rather than a package named for the api version?

api/core/v1alpha1/core.proto
```diff
@@ -0,0 +1,42 @@
...
+// Resource reference for an object
+message ObjectRef {
+
+    google.protobuf.StringValue api_group = 1;
+
+    google.protobuf.StringValue kind = 2;
+
+    string name = 3;
+
+    string namespace = 4;
```
ilackarms, line 23: (comments)

api/core/v1alpha1/core.proto
```diff
@@ -0,0 +1,42 @@
...
+    PROCESSING = 1;
+
+    // Invalid parameters supplied, will not continue.
+    INVALID = 2;
+
+    // Failed during processing.
+    FAILED = 3;
+
+    // Finished processing successfully.
+    ACCEPTED = 4;
```
ilackarms, line 41: i'm not sure about making `State` a core api in skv2 yet. i think it's worth discussing at least

api/generate.go
```diff
@@ -0,0 +1,60 @@
+package main
```
ilackarms, line 1: we already have some code generation in skv2 for multicluster; maybe this belongs alongside that?

api/multicluster/v1alpha1/cluster.proto
```diff
@@ -0,0 +1,29 @@
...
+option go_package = "github.com/solo-io/skv2/pkg/api/multicluster.solo.io/v1alpha1/types";
+
+
+
+/*
+    Representation of a Kubernetes cluster that has been registered.
+*/
+message KubernetesClusterSpec {
+    // name of the secret which contains the kubeconfig with information to connect to the remote cluster.
+    string secret = 1;
```
ilackarms, line 18: i'd like to allow the user to provide a secret_key here as well. we can override emtpy with a default

api/multicluster/v1alpha1/cluster.proto
```diff
@@ -0,0 +1,29 @@
...
+    // version of kubernetes
+    string version = 2;
+    // cloud provider, empty if unknown
+    string cloud = 3;
+
+    string region = 4;
+
+    string zone = 5;
```
ilackarms, line 26: are any of these other fields used anywhere? how are they intended to be consumed?

api/core/v1alpha1/core.proto
EItanya: Why no v1alpha1, using the versioning just allows us to make breaking changes to the API without having to make breaking changes to the code. As we support more APIs, and potentially multiple versions of those APIs, following kubernetes versioning for our APIs will become more and more useful.

api/multicluster/v1alpha1/cluster.proto
EItanya: What do you mean by a `secret_key`

api/generate.go
EItanya: We have templates to generate multi cluster reconcilers, but I don't think we have code to specifically generate code in this repo to be consumed by other repos, I may be wrong though.

api/multicluster/v1alpha1/cluster.proto
joekelley: I think he means the key of the kubeconfig entry in the secret.

```yaml
kind: Secret
metadata:
  ...
data:
  <cluster.secret_key | "some-default">: encoded_kubeconfig
```

PR #39, opened 2020-05-14
Remove Helm 2 crd-install hook

GrahamGoudeau:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/38

### DIFF
--- a/changelog/v0.3.0/remove-helm-2-hook.yaml
+++ b/changelog/v0.3.0/remove-helm-2-hook.yaml
@@ -1,0 +1,4 @@
+changelog:
+- type: BREAKING_CHANGE
+  description: Remove Helm 2 crd-install hook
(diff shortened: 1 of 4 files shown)


changelog/v0.3.0/remove-helm-2-hook.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+- type: BREAKING_CHANGE
```
ilackarms, line 2: i think we can just mark this as a FIX or non-user-facing as this feature isn't in use anywhere

GrahamGoudeau: Marked as non-user-facing since no Helm 2 clients can currently be consuming these charts anyway.

PR #43, opened 2020-05-29
Fix handling of finalized resources

joekelley:
Return early from Reconcile in a FinalizedReconciler in the event that our finalizer has already been processed.

This fixes a bug where finalized resources would be reconciled even though finalizers were cleared and a deletion timestamp was present.﻿


EItanya:
I discussed this with @joekelley and I approve of it

Issue #44, opened 2020-05-30
Add "set" generation for CRD types


harveyxia:
When dealing with collections of generated CRD types, it would be (extremely) helpful to have a basic set data structure. Leveraging the fact that k8s resources are uniquely identified by `name` and `namespace`, we can leverage the k8s [string set](https://github.com/kubernetes/apimachinery/blob/release-1.18/pkg/util/sets/string.go) in order to build sets for CRD types.

As discussed with @ilackarms, we'd also like to create directory structure to support contributions to skv2, as well as a process for graduating contributed features to core skv2 functionality if they prove valuable enough.

PR #45, opened 2020-05-30
Contrib set

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/44
resolves https://github.com/solo-io/skv2/issues/46


codegen/render/kube_types_renderer.go
```diff
@@ -23,6 +23,13 @@ type KubeCodeRenderer struct {
 	// the relative path to the api dir
 	// types will render in the package <module>/<apiRoot>/<group>/<version>
 	ApiRoot string
+
+	ContribTemplates ContribTemplates
+}
+
+type ContribTemplates struct {
```
EItanya, line 30: Why are these all being called `Contrib`? RenderMatchers seems more appropriate here.

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,114 @@
...
+import (
+    {{- if $.CustomTypesImportPath }}
+
+    . "{{ $.CustomTypesImportPath }}"
+    {{- end}}
+)
+
+// k8s resources are uniquely identified by their name and namespace
+func key(objectMeta v1.ObjectMeta) string {
+    return objectMeta.GetName() + "-" + objectMeta.GetNamespace()
```
EItanya, line 12: Any reason for using a `"-"` here. We commonly concatenate these values with a `"."` elsewhere.

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,114 @@
...
+}
+
+{{- range $resource := $.Resources }}
+
+{{- $kindLowerCamel := lower_camel $resource.Kind }}
+
+type {{ $resource.Kind }}Set interface {
+    Set() sets.String
```
EItanya, line 20:
```suggestion
    Keys() sets.String
```
`Set()` is slightly ambiguous, can we rename to `Keys()` or possibly `Ids()`

pkg/contrib/sets/string.go
```diff
@@ -0,0 +1,191 @@
+package sets
+
+import (
+	"reflect"
+	"sort"
+)
+
+/*
+	Copied from https://github.com/kubernetes/apimachinery/blob/release-1.18/pkg/util/sets/string.go
```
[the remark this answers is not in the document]
harveyxia: @ilackarms recommended it as a way to reduce external dependencies. I'm not opinionated here

codegen/render/kube_types_renderer.go
harveyxia: This struct is meant to house all contributed templates

pkg/contrib/sets/string.go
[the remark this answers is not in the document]
EItanya: I agree with the spirit of that, but we are already pulling in this library, so it feels unnecessary. I'd prefer to just use the kube one.

codegen/render/kube_types_renderer.go
EItanya: Gotcha, I think once something has been added to skv2 we should make it a first class citizen. If we keep adding things to the `Contrib` struct it will get unruly and confusing.

harveyxia: Wanted guidance on this very question—the code here was meant only as a placeholder in lieu of an actual structure for contributed templates

codegen/render/kube_types_renderer.go
```diff
@@ -73,13 +80,36 @@ func RenderApiTypes(grp Group) ([]OutFile, error) {
...
+			Path: "sets/set_matchers.go",
+		},
+	},
+}
+
+func (r KubeCodeRenderer) addContribTemplates(grp Group, templatesToRender inputTemplates) {
+	if grp.RenderContrib.Sets {
+		templatesToRender.add(r.ContribTemplates.Sets)
+	}
+	if grp.RenderContrib.Sets {
```
marcogschmidt, line 108:
```suggestion
	if grp.RenderContrib. SetMatchers {
```

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,114 @@
...
+    List() []*{{ $resource.Kind }}
+    Map() map[string]*{{ $resource.Kind }}
+    Insert({{ $kindLowerCamel }} ...*{{ $resource.Kind }})
+    Equal({{ $kindLowerCamel }}Set {{ $resource.Kind }}Set) bool
+    Has({{ $kindLowerCamel }} *{{ $resource.Kind }}) bool
+    Delete({{ $kindLowerCamel }} *{{ $resource.Kind }})
+    Union(set {{ $resource.Kind }}Set) {{ $resource.Kind }}Set
+    Difference(set {{ $resource.Kind }}Set) {{ $resource.Kind }}Set
+    Intersection(set {{ $resource.Kind }}Set) {{ $resource.Kind }}Set
```
marcogschmidt, line 29: Really looking forward to these!

pkg/contrib/sets/string.go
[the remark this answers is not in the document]
ilackarms: sorry, it was my understanding we were going to copy this and modify to support the `client.ObjectKey` as our key. if we're going with strings we don't need to copy this in

codegen/render/kube_types_renderer.go
harveyxia: thanks for the catch

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,114 @@
...
+func (s {{ $kindLowerCamel }}Set) List() []*{{ $resource.Kind }} {
+    var {{ $resource.Kind }}s []*{{ $resource.Kind }}
+    for _, key := range s.set.List() {
+        {{ $resource.Kind }}s = append({{ $resource.Kind }}s, s.mapping[key])
+    }
+    return {{ $resource.Kind }}s
+}
+
+func (s {{ $kindLowerCamel }}Set) Map() map[string]*{{ $resource.Kind }} {
+    return s.mapping
```
GrahamGoudeau, line 61: we probably don't want to export the mutable, internal data structure from here

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,114 @@
...
+    for _, {{ $resource.Kind }} := range {{ $resource.Kind }}s {
+        key := key({{ $resource.Kind }}.ObjectMeta)
+        set.Insert(key)
+        mapping[key] = {{ $resource.Kind }}
+    }
+    return &{{ $kindLowerCamel }}Set{set: set, mapping: mapping}
+}
+
+func (s {{ $kindLowerCamel }}Set) Set() sets.String {
+    return s.set
```
GrahamGoudeau, line 49: same comment here, this is a mutable internal datastructure, best practice is to copy it into a new thing before handing it out to callers

codegen/templates/contrib/sets/set_matchers.gotmpl
```diff
@@ -0,0 +1,39 @@
+package {{.Version}}
+
+import (
+    "fmt"
+
+    "github.com/onsi/gomega/types"
```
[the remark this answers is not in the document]
harveyxia: turns out we can just use vanilla gomema.Equal, so we don't need matchers at all

codegen/templates/contrib/sets/sets.gotmpl
```diff
@@ -0,0 +1,118 @@
...
+	for k, v := range s.mapping {
+		newMap[k] = v
+	}
+	return newMap
+}
+
+func (s {{ $kindLowerCamel }}Set) Insert(
+    {{ $resource.Kind }}s ...*{{ $resource.Kind }},
+) {
+    for _, {{ $resource.Kind }} := range {{ $resource.Kind }}s {
```
ilackarms, line 71: `for _,  {{ $kindLowerCamel }} `

Issue #47, opened 2020-06-08
helm chart generation is skipped when FilterTemplate func not provided


ilackarms:
caused by a bug in the logic surrounding the ability to exclude files from the skv2 generated helm chart.

PR #48, opened 2020-06-08
fix chartgen

ilackarms:
fixes an issue where the helm chart is not generated unless a `FilterTemplate` function is provided
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/47

### DIFF
--- a/codegen/render/chart_renderer.go
+++ b/codegen/render/chart_renderer.go
@@ -58,15 +58,16 @@ func (r ChartRenderer) Render(chart model.Chart) ([]OutFile, error) {
 
 	files = append(files, customFiles...)
 
-	var filteredFiles []OutFile
-	if chart.FilterTemplate != nil {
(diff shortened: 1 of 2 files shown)


codegen/render/chart_renderer.go
```diff
@@ -58,15 +58,16 @@ func (r ChartRenderer) Render(chart model.Chart) ([]OutFile, error) {
 
 	files = append(files, customFiles...)
 
-	var filteredFiles []OutFile
-	if chart.FilterTemplate != nil {
+	if chart.FilterTemplate == nil {
+		var filteredFiles []OutFile
 		for _, file := range files {
 			if chart.FilterTemplate(file.Path) {
```
marcogschmidt, line 64: Won't this always result in a nil pointer since `chart.FilterTemplate` is nil?

PR #59, opened 2020-06-18
Upgrades

ilackarms:
contains a set of changes which can be split out into separate PRs
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/60
resolves https://github.com/solo-io/skv2/issues/61
resolves https://github.com/solo-io/skv2/issues/62
resolves https://github.com/solo-io/skv2/issues/63
resolves https://github.com/solo-io/skv2/issues/64
resolves https://github.com/solo-io/skv2/issues/65


codegen/templates/code/types/clients.gotmpl
```diff
@@ -232,4 +232,26 @@ func (c *{{ $kindLowerCamel }}Client) Patch{{ $resource.Kind }}Status(ctx contex
     return c.client.Status().Patch(ctx, obj, patch, opts...)
 }
 
+// Provides {{ $resource.Kind }}Clients for multiple clusters.
+type Multicluster{{ $resource.Kind }}Client interface {
```
joekelley, line 236: 👏

api/core/v1/core.proto
```diff
@@ -28,6 +28,19 @@ message ObjectRef {
     string namespace = 4;
 }
 
+// Resource reference for an object
+message ClusterObjectRef {
```
ashleywang1, line 32: Nice!!! This will be super useful in gloo-hub! :)

contrib/pkg/sets/sets.go
```diff
@@ -1,36 +1,45 @@
...
+	Insert(resource ...ezkube.ResourceId)
 	Equal(resourceSet ResourceSet) bool
-	Has(resource metav1.Object) bool
-	Delete(resource metav1.Object)
+	Has(resource ezkube.ResourceId) bool
+	Delete(resource ezkube.ResourceId)
 	Union(set ResourceSet) ResourceSet
 	Difference(set ResourceSet) ResourceSet
 	Intersection(set ResourceSet) ResourceSet
+	Find(resoruceType, id ezkube.ResourceId) (ezkube.ResourceId, error)
```
harveyxia, line 31:
```suggestion
	Find(resourceType, id ezkube.ResourceId) (ezkube.ResourceId, error)
```

contrib/pkg/sets/sets.go
```diff
@@ -91,7 +112,7 @@ func (s resourceSet) Union(set ResourceSet) ResourceSet {
 
 func (s resourceSet) Difference(set ResourceSet) ResourceSet {
 	newSet := s.set.Difference(set.Keys())
```
harveyxia, line 114: Do we need a lock for Difference and Intersection?

codegen/templates/code/types/type_helpers.gotmpl
```diff
@@ -0,0 +1,16 @@
...
+    metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+    {{- if $.CustomTypesImportPath }}
+    . "{{ $.CustomTypesImportPath }}"
+    {{- end}}
+)
+
+{{- range $resource := $.Resources }}
+
+// {{ $resource.Kind }}Slice represents a slice of *{{ $resource.Kind }}
+type {{ $resource.Kind }}Slice []*{{ $resource.Kind }}
```
EItanya, line 14: What do you think about putting a Sort method on this, or possibly on the sets?

contrib/pkg/sets/sets.go
```diff
@@ -101,10 +122,26 @@ func (s resourceSet) Difference(set ResourceSet) ResourceSet {
...
-	var newResources []metav1.Object
+	var newResources []ezkube.ResourceId
 	for key, _ := range newSet {
 		val, _ := s.mapping[key]
 		newResources = append(newResources, val)
 	}
 	return NewResourceSet(newResources...)
 }
+
+func (s resourceSet) Find(
```
EItanya, line 133: This feels strange to me, why are we passing a `resourceType` and an `Id` here of the same type?

pkg/multicluster/watcher.go
```diff
@@ -19,3 +19,9 @@ type ManagerSet interface {
 	// Cluster returns a manager for the given cluster, or an error if one does not exist.
 	Cluster(cluster string) (manager.Manager, error)
 }
+
+// ManagerSet maintains a manager for every cluster in the system.
+type ClusterSet interface {
+	// List the clusters currently known to the set
```
EItanya, line 25:
```suggestion
	// List the clusters (sorted) currently known to the set
```

contrib/codegen/templates/sets/sets.gotmpl
```diff
@@ -103,4 +104,14 @@ func (s {{ $kindLowerCamel }}Set) Intersection(set {{ $resource.Kind }}Set) {{ $
     return New{{ $resource.Kind }}Set({{ $kindLowerCamel }}List...)
 }
 
+
+func (s {{ $kindLowerCamel }}Set) Find(id ezkube.ResourceId) (*{{ $import_prefix }}.{{ $resource.Kind }}, error) {
```
EItanya, line 108: `id ezkube.ResourceId` doesn't feel like the right arg for this function. Most times this function will most likely be called with the user only having the string information. My recommendation here would be to either, have a simple companion struct which implements `ezkube.ResourceId`
for instance:
```
type ResourceKey struct {
    Name string
    Namespace string
    Cluster string
}
// implemented methods
```
This way find can be called similar to how other Gets/Reads work throughout sk and skv2

contrib/pkg/sets/sets.go
```diff
@@ -1,36 +1,45 @@
...
 
+var NotFoundErr = func(resourceType ezkube.ResourceId, id ezkube.ResourceId) error {
+	return eris.Errorf("%T with id %v not found", resourceType, Key(id))
+}
+
 // k8s resources are uniquely identified by their name and namespace
-func Key(objectMeta metav1.Object) string {
-	return objectMeta.GetName() + "." + objectMeta.GetNamespace()
+func Key(id ezkube.ResourceId) string {
+	return id.GetName() + "." + id.GetNamespace() + "." + id.GetClusterName()
```
EItanya, line 17: This should most likely omit the ClusterName if it is not available

contrib/codegen/templates/sets/sets.gotmpl
ilackarms: you can use a `ClusterObjectRef` as the argument for Find, or an existing object. I prefer the interface because of the flexibility it offers

contrib/pkg/sets/sets.go
ilackarms: i was thinking the same, but then the same could apply to Namespace (since it's empty on cluster scoped objects). my preference is to leave it the way it is, as it's only used as a key in a Set

codegen/templates/code/types/type_helpers.gotmpl
ilackarms: i don't follow; set is a map

and i'm happy to introduce a sort method here but i didn't need one. i think this PR is a bit large enough, but happy to approve it in a follow up

PR #69, opened 2020-06-25
expose local domain override on default registrant

joekelley:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/76


pkg/multicluster/register/helpers.go
```diff
@@ -49,7 +49,7 @@ func RegisterClusterFromConfig(
 	Meant to be used in tandem with RegisterClusterFromConfig above.
 	They are exposed separately so the `Registrant` may be mocked for the function above.
 */
-func DefaultRegistrant(context string) (ClusterRegistrant, error) {
+func DefaultRegistrant(context, localDomainOverride string) (ClusterRegistrant, error) {
```
ilackarms, line 52: can we comment the purpose of `localDomainOverride` / what it should be set to?

changelog/v0.5.1/expose-local-domain-override.yaml
```diff
@@ -0,0 +1,3 @@
+changelog:
+  - type: NON_USER_FACING
```
ilackarms, line 2: i think this is technically a breaking change..

PR #71, opened 2020-06-26
input snapshot

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/70


contrib/custom_templates.go
```diff
@@ -54,3 +56,25 @@ var ClientProviders = func() model.CustomTemplates {
...
 }()
+
+/*
+InputSnapshot custom template
+*/
+const (
+	InputSnapshotCustomTemplatePath = "input/input_snapshot.gotmpl"
+)
+
+var InputSnapshot = func(outputFilename, groupModule string, selectFromGroups []model.Group, resourcesToSelect map[schema.GroupVersion][]string) model.CustomTemplates {
```
EItanya, line 67: Since this is the top-level function users will be interacting with, can we include more extensive documentation here as well.

Issue #91, opened 2020-07-06
make it easier to construct object refs


ilackarms:
it's very rote/repetitive work to always construct manually.

Issue #90, opened 2020-07-06
implement interface for local object references


ilackarms:
when objects are known to exist in a local cluster, it is burdensome to require references to implement `GetClusterName()`.

we should amend the current ezkube.ResourceId to support both cases (cluster- and locally scoped references).

PR #92, opened 2020-07-06
Local resource object refs

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/90
resolves https://github.com/solo-io/skv2/issues/91


pkg/utils/refutils/object_ref.go
```diff
@@ -0,0 +1,21 @@
+package refutils
```
EItanya, line 1: Do we really need a whole knew package for this, why not just put it in `ezkube`?

PR #97, opened 2020-07-08
allow overriding proto package for proto-based resources

ilackarms:
When using proto-based types for the `Spec` and `Status` fields of a resource, we currently require that the messages for these types are defined in a proto package that matches the group name of the resource.

This change adds a new `ProtoPackage` option that can be used to override the name of the proto packages we inspect when looking for the messages. If the options in unspecified, it defaults to the group name of the resource.

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/98

### DIFF
--- a/codegen/render/proto_types_renderer.go
+++ b/codegen/render/proto_types_renderer.go
@@ -23,7 +23,6 @@ type ProtoCodeRenderer struct {
 func RenderProtoTypes(grp Group) ([]OutFile, error) {
 	defaultKubeCodeRenderer := ProtoCodeRenderer{
 		templateRenderer: DefaultTemplateRenderer,
-		GoModule:         grp.Module,
 		ApiRoot:          grp.ApiRoot,
 	}
 
@@ -33,7 +32,7 @@ func RenderProtoTypes(grp Group) ([]OutFile, error) {
 func (r ProtoCodeRenderer) RenderProtoHelpers(grp Group) ([]OutFile, error) {
 
 	// only render proto helpers for proto groups in the current module
-	if !grp.HasProtos() || grp.Module != r.GoModule {
+	if !grp.HasProtos() {
 		return nil, nil
 	}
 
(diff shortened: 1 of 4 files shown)


codegen/render/proto_types_renderer.go
```diff
@@ -33,7 +33,7 @@ func RenderProtoTypes(grp Group) ([]OutFile, error) {
 func (r ProtoCodeRenderer) RenderProtoHelpers(grp Group) ([]OutFile, error) {
 
 	// only render proto helpers for proto groups in the current module
-	if !grp.HasProtos() || grp.Module != r.GoModule {
```
marcogschmidt, line 36: The second clause of this statement would always evaluate to `false`.

EItanya: Are you sure? even in other projects. I don't want downstream projects broken based on this change

marcogschmidt: In this project yes, since `Module` of the `ProtoCodeRenderer` is set to the one of the `Group` when we construct it (see [here](https://github.com/solo-io/skv2/blob/a3079f6f4f9e0263b004d0ee32cf22330c863574/codegen/render/proto_types_renderer.go#L26).

Hadn't considered that the type is exported, though. Is it directly used in other projects? I don't think it is usable outside of this package: there is no way to construct a valid instance since the embedded `templateRenderer` is private and is needed for it to work AFAICT:

```go
type ProtoCodeRenderer struct {
	templateRenderer

	// the go module of the project
	GoModule string

	// the relative path to the api dir
	// types will render in the package <module>/<apiRoot>/<group>/<version>
	ApiRoot string
}
```

I could:
- just revert this change, or
- export the `templateRenderer` type, or
- make the `ProtoCodeRenderer` private

Don't feel strongly either way, wdyt?

EItanya: Fine as is, thanks for the explination!

Issue #100, opened 2020-07-12
expose KubernetesCluster codegen goup


ilackarms:
for use in Top-Level templates in other projects

Issue #99, opened 2020-07-12
generate KuberentesCluster CR on cluster registration


ilackarms:
note that in order to populate the ClusterDomain field for this CR, we'll need to rename the currently (mis-named) ClusterDomainOverride -> APIServerAddress

PR #101, opened 2020-07-12
generate kubecluster crd on cluster registration

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/102
resolves https://github.com/solo-io/skv2/issues/99
resolves https://github.com/solo-io/skv2/issues/100


contrib/pkg/input/input_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+		err = r.singleClusterReconcileFunc(key.(ezkube.ResourceId))
+	case r.multiClusterReconcileFunc != nil:
+		err = r.multiClusterReconcileFunc(key.(ezkube.ClusterResourceId))
+	}
+
+	if err == nil {
+		r.queue.Forget(key)
+	} else {
+		contextutils.LoggerFrom(r.ctx).Errorw("encountered error reconciling state; retrying", "error", err)
+		r.queue.AddRateLimited(key)
```
EItanya, line 131: Can we also allow for a requeue with no error. In `gloo-fed` we have already encountered a situation where returning an error creates an absolute mess in the logs for no reason, because the only return option is an error, and error logging is REALLY verbose.

PR #107, opened 2020-07-14
Add typed object refs.

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/106


api/core/v1/core.proto
```diff
@@ -22,7 +22,7 @@ message ObjectRef {
     string namespace = 2;
 }
 
-// Resource reference for an object
+// Resource reference for a cluster-scoped object
```
ilackarms, line 25: i think of this as a cross-cluster-scoped reference, wdyt?

Issue #134, opened 2020-09-01
Extend cluster registration to accept in memory kubeconfigs


harveyxia:
In certain contexts, such as automated EKS cluster discovery, we need the ability to register a cluster given an in-memory kubeconfig (i.e. `clientcmd.ClientConfig`). We should extend the registration logic to allow for this option.

PR #135, opened 2020-09-02
Cluster registration clientconfig

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/134


pkg/multicluster/register/helpers.go
```diff
@@ -248,27 +226,6 @@ func defaultRegistrant(cfg *rest.Config, apiServerAddress string) (ClusterRegist
 	return registrant, nil
 }
 
-// Attempts to load a Client KubeConfig from a default list of sources.
-func getClientConfigWithContext(masterURL, kubeCfgPath, context string) (clientcmd.ClientConfig, error) {
```
ilackarms, line 252: can we expose this as a util function? seems like all the consuming projects would need to re-implement otherwise

harveyxia: This is already [available here](https://github.com/solo-io/skv2/blob/b3439f5151a08654a765bce3111d2fdcdfa5009e/pkg/multicluster/kubeconfig/loader.go#L23)

pkg/multicluster/register/helpers.go
```diff
@@ -23,22 +21,15 @@ import (
...
-	RemoteKubeCfgPath string
+	// Remote kubeconfig
+	RemoteKubeCfg clientcmd.ClientConfig
 
-	// override the context to use from the remote kubeconfig
-	// if unset, use current context
-	RemoteKubeContext string
+	// Remote context name
+	// We need to explicitly pass this because of this open issue: https://github.com/kubernetes/client-go/issues/735
+	RemoteCtx string
```
EItanya, line 32: We should probably allow overriding the local context as well

PR #137, opened 2020-09-02
support deepcopy generation for non-proto types

ilackarms:


codegen/render/kube_types_renderer.go
```diff
@@ -73,7 +68,7 @@ var ControllerTemplates = inputTemplates{
 func RenderApiTypes(grp Group) ([]OutFile, error) {
 	defaultKubeCodeRenderer := KubeCodeRenderer{
 		templateRenderer:    DefaultTemplateRenderer,
-		TypesTemplates:      TypesTemplates(grp.Generators.HasDeepcopy()),
```
EItanya, line 76: So after this change is `HasDeepcopy()` not respected anymore?

ilackarms: no projects currently leverage the functionality, i think we ought to just get rid of it

EItanya: Well then remove the variable altogether
We shouldn't leave something unimplemented

EItanya (approved):

Issue #140, opened 2020-09-04
input snapshot: handle case when crds not registered


ilackarms:
currently, if an input snapshot includes CRDs that are not registered, building the snapshot will fail with a `NoKindMatchError`.

in certain cases, it is valid to attempt to build a snapshot consisting of CRDs when it is not known ahead of time whether the CRD is registered in the cluster. this is the case for Service Mesh Hub, which does not know whether AppMesh CRDs have been installed to a registered cluster prior to Discovery.

i will be adding an option to the input.BuildOptions which are passed to the input builder to allow ignoring certain types of errors

PR #141, opened 2020-09-08
input Snapshot builder: check CRD registration

ilackarms:
exposes an option to verify the existence of a CRD before attempting to read its instances from a cluster.

note: requires that the CRD type is registered to the scheme used by the provided `client`
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/140

### DIFF
--- a/contrib/codegen/templates/input/input_reconciler.gotmpl
+++ b/contrib/codegen/templates/input/input_reconciler.gotmpl
@@ -17,6 +17,8 @@ package {{ package }}
 import (
     "context"
 
+    "github.com/solo-io/skv2/pkg/verifier"
+
     "github.com/solo-io/skv2/contrib/pkg/input"
     sk_core_v1 "github.com/solo-io/skv2/pkg/api/core.skv2.solo.io/v1"
     "github.com/solo-io/skv2/pkg/multicluster"
@@ -57,6 +59,21 @@ type multiClusterReconcilerImpl struct {
     base input.MultiClusterReconciler
 }
 
+// Options for reconcileing a snapshot
+type ReconcileOptions struct {
+{{/* generate reconcile options here */}}
+{{- range $group := $groups }}
+{{ $client_import_prefix := group_import_name $group }}
+{{- range $resource := $group.Resources }}
+{{- $kindPlural := pluralize $resource.Kind }}
+{{- $kindLowerCamel := lower_camel $resource.Kind }}
+{{- $kindLowerCamelPlural := pluralize $kindLowerCamel }}
+    // Options for reconciling {{ $kindPlural }}
+    {{ $kindPlural }} reconcile.Options
+{{- end }}
+{{- end }}
+}
+
 // register the reconcile func with the cluster watcher
 // the reconcileInterval, if greater than 0, will limit the number of reconciles
 // to one per interval.
@@ -65,6 +82,7 @@ func RegisterMultiClusterReconciler(
         clusters multicluster.ClusterWatcher,
         reconcileFunc input.MultiClusterReconcileFunc,
         reconcileInterval time.Duration,
+        options ReconcileOptions,
         predicates ...predicate.Predicate,
 ) {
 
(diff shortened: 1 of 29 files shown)


pkg/multicluster/watcher.go
```diff
@@ -28,3 +28,10 @@ type ClusterSet interface {
 	// List the clusters (sorted) currently known to the set
 	ListClusters() []string
 }
+
+// the multicluster Interface provides a handle to interacting with multiple clusters.
+// the multicluster watcher implements this interface.
+type Interface interface {
```
EItanya, line 34: Is `ClusterSet` specifically missing from this interface? and If so do we still wanna check that the `clusterWatcher` implements it?

contrib/codegen/templates/input/input_snapshot.gotmpl
```diff
@@ -201,43 +211,61 @@ type BuildOptions struct {
...
 {{- end }}
 }
 
-// build a snapshot from resources across multiple clusters
-type multiClusterBuilder struct {
-    clusters     multicluster.ClusterSet
-{{/* generate client fields here */}}
+type CrdCheckOption int
+
+const (
```
EItanya, line 221: nit: any reason we are generating this const every time?

contrib/codegen/templates/input/input_snapshot.gotmpl
```diff
@@ -415,3 +478,22 @@ func (b *singleClusterBuilder) insert{{ $kindPlural }}(ctx context.Context, {{ $
 
 {{- end }}
 {{- end }}
+
+func verifyCrdRegistered(ctx context.Context, cfg *rest.Config, group, version, kind string) (bool, error) {
```
EItanya, line 482: Same nit as above, why generate this for every snapshot?

pkg/multicluster/watcher.go
ilackarms: `Watcher` already includes `ClusterSet` (so it's actually a compile error to include them both here)

contrib/codegen/templates/input/input_snapshot.gotmpl
ilackarms: no longer relevant, the code has been moved

contrib/codegen/templates/input/input_reconciler.gotmpl
```diff
@@ -17,6 +17,8 @@ package {{ package }}
 import (
     "context"
 
+    "github.com/solo-io/skv2/pkg/verifier"
+
```
EItanya, line 21: nit: extra new line

contrib/codegen/templates/input/input_snapshot.gotmpl
```diff
@@ -201,47 +214,38 @@ type BuildOptions struct {
...
+// Options for reading resources of a given type
+type ResourceBuildOptions struct {
+{{/* generate build options here */}}
+    // List options for composing a snapshot from a resource type
+    ListOptions []client.ListOption
+
+    // If provided, ensure the resource has been verified before adding it to snapshots
+    Verifier verifier.ServerResourceVerifier
+}
+
```
EItanya, line 231: Any reason to generate this many times rather than have it live one place?

pkg/reconcile/runner.go
```diff
@@ -80,6 +97,15 @@ func (r *runner) RunReconciler(ctx context.Context, reconciler Reconciler, predi
 	if err != nil {
 		return err
 	}
+
+	if r.options.Verifier != nil {
+		if resourceRegistered, err := r.options.Verifier.VerifyServerResource(r.cluster, r.mgr.GetConfig(), gvk); err != nil {
+			return err
+		} else if !resourceRegistered {
+			return nil
```
EItanya, line 105: Should we log here to explain what happened, or maybe return a different error, because it could be confusing for nothing to happen

pkg/verifier/verifier.go
```diff
@@ -0,0 +1,147 @@
...
+	return &verifier{
+		ctx:                      ctx,
+		options:                  options,
+		verifiedClusterResources: map[string]map[schema.GroupVersionKind]bool{},
+	}
+}
+
+// Verify whether the API server for the given rest.Config supports the resource with the given GVK.
+func (v *verifier) VerifyServerResource(cluster string, cfg *rest.Config, gvk schema.GroupVersionKind) (bool, error) {
+	v.lock.RLock()
```
EItanya, line 75: The use of locks here is a bit hard to reason about, could we possibly move the map reads and writes to their own functions for clarity?

pkg/reconcile/runner.go
ilackarms: if the user sets the verify option to `Warn` an warning will be logged

contrib/codegen/templates/input/input_snapshot.gotmpl
ilackarms: it makes it easier for the user to write, by only needing to import one package

contrib/codegen/templates/input/input_reconciler.gotmpl
ilackarms: there's a loose convention i've been follow as far as ordering import statements: when a new import gets added to a file, goimports tends to sort it in this way. the spacing actually gives a little bit of historical context as to when an import was added to a file. no strong objection to changing here, but i have a slight preference to leave as-is. let me know if u want the change and i dont mind to do it

EItanya (approved):

Issue #142, opened 2020-09-08
Extend KubernetesCluster status with namespace and policy rules


harveyxia:
Extend the KubernetesCluster status with the namespace and policy rules that the cluster was registered with, so that the cluster-discovery pod can know when to "resync" (i.e. reregister) the cluster with updated settings (e.g. namespace and/or policy rules)

PR #143, opened 2020-09-08
Extend KubernetesCluster status with registration settings

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/142


api/multicluster/v1alpha1/cluster.proto
```diff
@@ -56,4 +56,40 @@ message KubernetesClusterStatus {
     // List of statuses about the kubernetes cluster.
     // This list allows for multiple applications/pods to record their connection status.
     repeated core.skv2.solo.io.Status status = 1;
+
+    // The namespace on in which cluster registration resources are created.
+    string namespace = 2;
+
+    // JSON representation of the set of PolicyRules to attach to the newly created ClusterRole when registering clusters.
```
ilackarms, line 63: nit: fix comment ( it's no longer json)

harveyxia:
Closing in favor of the cluster registration solution proposed in https://github.com/solo-io/skv2/issues/144

pkg/multicluster/register/helpers.go
```diff
@@ -70,6 +70,9 @@ type RegistrationOptions struct {
 	// List of cluster roles which will be bound to by the created cluster role bindings
 	// The ClusterRoles upserted from the above list will automatically appended to the list
 	ClusterRoleBindings []client.ObjectKey
+
+	// Set of labels to include on the KubernetesCluster resource.
+	KubernetesClusterLabels map[string]string
```
ilackarms, line 75: wondering if this shouldn't apply to all output resources? and then `ResourceLabels`?

harveyxia: makes sense

pkg/multicluster/register/helpers.go
```diff
@@ -168,7 +192,14 @@ func RegisterProviderClusterFromConfig(
 	opts RbacOptions,
 	registrant ClusterRegistrant,
 	providerInfo *v1alpha1.KubernetesClusterSpec_ProviderInfo,
+	kubeClusterLabels map[string]string,
+	policyRules []*v1alpha1.PolicyRule,
```
ilackarms, line 196: given the number of options we're growing to here, i think it makes sense to encapsulate these in a struct. wdyt?

pkg/multicluster/register/registrant.go
```diff
@@ -289,7 +301,7 @@ func (c *clusterRegistrant) RegisterClusterWithToken(
 	token string,
 	opts Options,
 ) error {
-	return c.RegisterProviderClusterWithToken(ctx, masterClusterCfg, remoteClientCfg, token, opts, nil)
+	return c.RegisterProviderClusterWithToken(ctx, masterClusterCfg, remoteClientCfg, token, opts, nil, nil, nil)
```
ilackarms, line 304: what's with the nils? just curious. it would help to use a struct as a param here, so the strut fields could be named and commented with a bit of insight into what's optional

pkg/multicluster/register/interfaces.go
```diff
@@ -148,15 +155,15 @@ type ClusterRegistrant interface {
 	) error
 
 	/*
-		Same functionality as RegisterClusterWithToken but supply extra ProviderInfo metadata.
+		Same functionality as RegisterClusterWithToken but supply extra ProviderInfo metadata and registration settings (namespace and policyRules).
```
joekelley, line 158: nit: stale comment? maybe:
```suggestion
		Same functionality as RegisterClusterWithToken but supply extra RegistrationMetadata.
```

Since we already have registration opts, do we also need registration metadata? Seems to me like we could nest registation meta within the Options struct and remove this provider-specific function altogether.

harveyxia: Great suggestion, see the latest commit for the consolidated struct, a lot cleaner this way imo.

pkg/multicluster/register/interfaces.go
```diff
@@ -45,11 +46,25 @@ type Options struct {
...
+// Optional additional metadata to persist to registration output resources.
+type RegistrationMetadata struct {
+	// Metadata about the provider for cloud hosted k8s clusters.
+	ProviderInfo *v1alpha1.KubernetesClusterSpec_ProviderInfo
 
+	// Labels to add to registration output resources (KubernetesCluster and Secret).
+	ResourceLabels map[string]string
+
+	// The set of PolicyRules for the cluster roles created on the remote cluster upon registration.
+	ClusterRolePolicyRules []*v1alpha1.PolicyRule
```
ilackarms, line 64: are they always cluster roles? this wasn't clear to me

pkg/multicluster/register/interfaces.go
```diff
@@ -30,27 +30,41 @@ type Options struct {
...
 	RemoteNamespace string
 
 	// The Cluster Domain used by the Kubernetes DNS Service.
 	// Defaults to 'cluster.local'
 	// Read more: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/
 	ClusterDomain string
+
+	RegistrationMetadata RegistrationMetadata
+
+	RbacOptions RbacOptions
```
ilackarms, line 52: nice. much better than the previous structure of Options/RbacOptions

pkg/multicluster/register/interfaces.go
ilackarms: do you think it makes sense to put non-cluster rbac here?

pkg/multicluster/register/registrant.go
```diff
@@ -362,8 +379,8 @@ func (c *clusterRegistrant) DeregisterCluster(
 		return err
 	}
 
-	kcSecretObjMeta := kubeconfig.SecretObjMeta(opts.Namespace, opts.ClusterName)
-	kubeClusterObjMeta := kubeClusterObjMeta(kcSecretObjMeta.Name, kcSecretObjMeta.Namespace)
+	kcSecretObjMeta := kubeconfig.SecretObjMeta(opts.Namespace, opts.ClusterName, nil)
+	kubeClusterObjMeta := kubeClusterObjMeta(kcSecretObjMeta.Name, kcSecretObjMeta.Namespace, nil)
```
ilackarms, line 383: why `nil`?

harveyxia: This constructs object metas for the purposes of deregistration, so we don't need the resource labels, only the name/namespace

pkg/multicluster/register/interfaces.go
harveyxia: It looks like no project is registering with Roles, so I opted not to include that here

harveyxia: Discussed offline, there's reason to include this logic (we anticipate registering with Roles to become a requested use case in the coming months)

ilackarms (approved):

Issue #149, opened 2020-09-24
sets.Key() - We should impose a restriction that the cluster name cannot have a '.' in them.


ashleywang1:
In some places (smh apiserver) we need to convert back from a key generate by sets.Key() into an objectRef.

This restriction allows us to determine, for sure, what the name, namespace and cluster is.

It would also be nice to have a function that reverses the sets.Key() operation.

PR #150, opened 2020-09-28
Cluster name restriction

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/149


harveyxia:
Can we add a call to `validate` right after [this line](https://github.com/solo-io/skv2/blob/2965fd79e360e1175395d8bb3f941f6cee18b06c/pkg/multicluster/register/helpers.go#L172)? This is the entry point that SMH uses

harveyxia (approved):

PR #154, opened 2020-10-08
allow specifying kubeconfig path

harveyxia:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/153

### DIFF
--- a/pkg/multicluster/kubeconfig/loader.go
+++ b/pkg/multicluster/kubeconfig/loader.go
@@ -11,39 +11,50 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
-// Fetch ClientConfig for environment in which this is invoked, override the API Server URL and current context.
-// Lifted from https://github.com/kubernetes-sigs/controller-runtime/blob/cb7f85860a8cde7259b35bb84af1fdcb02c098f2/pkg/client/config/config.go#L135
-func GetClientConfigWithContext(apiServerUrl, kubeContext string) (clientcmd.ClientConfig, error) {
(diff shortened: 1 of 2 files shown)


harveyxia:
waiting for e2e tests to pass on the dependent PR, https://github.com/solo-io/service-mesh-hub/pull/999

pkg/multicluster/kubeconfig/loader.go
```diff
@@ -11,39 +11,50 @@ import (
...
 		return nil, err
 	}
 	return clientConfig.ClientConfig()
 }
 
-// Fetch raw Config for environment in which this is invoked, override the API Server URL and current context.
-func GetRawConfigWithContext(apiServerUrl, kubeContext string) (clientcmdapi.Config, error) {
-	clientConfig, err := GetClientConfigWithContext(apiServerUrl, kubeContext)
+// Fetch raw Config for environment in which this is invoked, override the API Server URL and current context if specified.
+func GetRawConfigWithContext(kubeConfigPath, kubeContext, apiServerUrl string) (clientcmdapi.Config, error) {
```
joekelley, line 56: WDYT about deprecating or getting rid of these helpers in favor of `GetClientConfigWithContext`? On the one hand it's nice to have a function signature we can quickly reference to find the type of kubeconfig we need, but on the other these create an additional maintenance burden on top of GetClientConfigWithContext.

harveyxia: I could go either way. I lean slightly towards keeping this convenience helpers since I don't expect we'll ever have to update them

joekelley: works for me! I think having these will help folks who are less familiar with the various kube config types find what they need a bit quicker.

joekelley (approved):

Issue #172, opened 2020-11-23
Add GVK method to generated Go structs


ryantking:
Currently, getting the GVK for is a bit convoluted requiring the scheme and some helper functions.

A GVK method can easily be autogenerated.

PR #173, opened 2020-11-23
Add GVK method to generated go structs

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/172

### DIFF
--- a/codegen/templates/code/types/types.gotmpl
+++ b/codegen/templates/code/types/types.gotmpl
@@ -38,8 +38,8 @@ type {{ $resource.Kind }} struct {
 func ({{ $resource.Kind}})  GVK() schema.GroupVersionKind {
 	return schema.GroupVersionKind{
 		Group: "{{ $resource.Group.Group }}",
-        Version: "{{ $resource.Version }}",
-        Kind: "{{ $resource.Kind }}",
+		Version: "{{ $resource.Version }}",
+		Kind: "{{ $resource.Kind }}",
     }
 }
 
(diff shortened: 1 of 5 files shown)


codegen/test/api/things.test.io/v1/types.go
```diff
@@ -21,6 +22,15 @@ type Paint struct {
 	Status PaintStatus `json:"status,omitempty"`
 }
 
+// GVK returns the GroupVersionKind associated with the resource type.
+func (Paint) GVK() schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group:   "things.test.io/v1",
```
ilackarms, line 28: these groups aren't generating correctly, should be `things.test.io`

codegen/templates/code/types/types.gotmpl
```diff
@@ -32,6 +34,15 @@ type {{ $resource.Kind }} struct {
 {{- end }}
 }
 
+// GVK returns the GroupVersionKind associated with the resource type.
+func ({{ $resource.Kind}})  GVK() schema.GroupVersionKind {
+	return schema.GroupVersionKind{
+		Group: "{{ $resource.Group }}",
```
ilackarms, line 40: i tihink you need resource.Group.Group here

codegen/test/api/things.test.io/v1/types.go
ryantking: Should this file be in VCS in general? It looks like its only created via tests, not codegen.

ilackarms (approved):

Issue #176, opened 2020-12-02
we should respect `--api-server-address` regardless of whether we use kind


ilackarms:
currently cluster registration only allows user to override the apiserver address in the registrant's kubeconfig if the cluster is a `kind` cluster. this requirement is unncessary/invalid; there are many potential use cases in which the address of the apiserver is different between the user's machine and the management cluster.

we should remove the `kind` check here and always respect `api-server-address` flag:

https://github.com/solo-io/skv2/blob/3796bc6c1d912c9c896f60b11107bd42ff1732d5/pkg/multicluster/register/registrant.go#L642

ashleywang1:
I'm gonna roll this in to a couple other changes I'm making. I ran into the following error:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x37f1796]

goroutine 1 [running]:
github.com/solo-io/skv2/pkg/multicluster/register.RegistrationOptions.initialize(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ffeefbff807, 0x14, 0x7ffeefbff7c0, 0xa, ...)
	/Users/ashwang/go/pkg/mod/github.com/solo-io/skv2@v0.13.5/pkg/multicluster/register/helpers.go:139 +0xa6
github.com/solo-io/skv2/pkg/multicluster/register.RegistrationOptions.RegisterProviderCluster(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ffeefbff807, 0x14, 0x7ffeefbff7c0, 0xa, ...)
	/Users/ashwang/go/pkg/mod/github.com/solo-io/skv2@v0.13.5/pkg/multicluster/register/helpers.go:104 +0xcb
github.com/solo-io/skv2/pkg/multicluster/register.RegistrationOptions.RegisterCluster(0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x7ffeefbff807, 0x14, 0x7ffeefbff7c0, 0xa, ...)
	/Users/ashwang/go/pkg/mod/github.com/solo-io/skv2@v0.13.5/pkg/multicluster/register/helpers.go:92 +0x8c
github.com/solo-io/gloo/projects/gloo/cli/pkg/cmd/federation/register.Register(0xc000bef500, 0x0, 0x0)
[...]

PR #186, opened 2020-12-14
Pass in templateContents string instead of a templatePath string to SnapshotTemplateParameters ConstructTemplate function.

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/184
resolves https://github.com/solo-io/skv2/issues/176

### DIFF
--- a/changelog/v0.15.2/custom-templates.yaml
+++ b/changelog/v0.15.2/custom-templates.yaml
@@ -2,7 +2,7 @@ changelog:
 - type: FIX
   description: >
     Make SnapshotTemplateParameters ConstructTemplate take in a params templateContents string instead
-    of a templatePath string.
+    of a templatePath string, as well as a mockgenDirective bool for additional customization.
   issueLink: https://github.com/solo-io/skv2/issues/184
 - type: NEW_FEATURE
   description: >
--- a/changelog/v0.15.2/custom-templates.yaml
+++ b/changelog/v0.15.2/custom-templates.yaml
@@ -1,10 +1,10 @@
 changelog:
-- type: NON_USER_FACING
+- type: FIX
   description: >
     Make SnapshotTemplateParameters ConstructTemplate take in a params templateContents string instead
     of a templatePath string.
   issueLink: https://github.com/solo-io/skv2/issues/184
-- type: NON_USER_FACING
+- type: NEW_FEATURE
   description: >
     Change hackClusterConfigForLocalTestingInKIND to just hackClusterConfigForLocalTesting, and remove
     the requirement that it be a KIND cluster. Respect the `--api-server-address` flag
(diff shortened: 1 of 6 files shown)


changelog/v0.15.2/custom-templates.yaml
```diff
@@ -0,0 +1,12 @@
+changelog:
```
joekelley, line 1: I think these changes are fixes or new features

joekelley: or at least the registration change

pkg/multicluster/register/registrant.go
```diff
@@ -630,17 +628,15 @@ func (c *clusterRegistrant) getTokenForSa(
...
-	remoteContextName string,
 ) error {
 	serverUrl, err := url.Parse(remoteCluster.Server)
 	if err != nil {
 		return err
 	}
 
-	if strings.HasPrefix(remoteContextName, "kind-") &&
-		(serverUrl.Hostname() == "127.0.0.1" || serverUrl.Hostname() == "localhost") &&
+	if (serverUrl.Hostname() == "127.0.0.1" || serverUrl.Hostname() == "localhost") &&
```
ilackarms, line 639: i think we should modify this so that we always respect `c.localAPIServerAddress` if it is set; there are use cases where the address a user uses to communicate with kube from their local machine may differ from the one used by pods running in the cluster.

so i think what that would mean is to separate out the following:

```go
if c.localAPIServerAddress {
	if host, newPort, err := net.SplitHostPort(c.localAPIServerAddress); err == nil {
			c.localAPIServerAddress = host
			port = newPort
		}
	}
	remoteCluster.Server = fmt.Sprintf("https://%s:%s", c.localAPIServerAddress, port)
}
```

as far as the other changes we do, we can comntiue to put them behind the check for 
`serverUrl.Hostname() == "127.0.0.1" || serverUrl.Hostname() == "localhost"`, though i think we might want to consider exposing flags for these as well (as there are other situations where a user may want to configure custom options for TLS (outside of using Kind/minikube)

pkg/multicluster/register/registrant.go
```diff
@@ -620,38 +622,50 @@ func (c *clusterRegistrant) getTokenForSa(
...
+}
+
+// if:
+//   * the --local-cluster-domain-override flag is populated with a value
+//
+// then rewrite the server config to communicate over the value of
+// `--local-cluster-domain-override`, which resolves to the host machine of docker.
+// There are use cases where the address a user uses to communicate with kubernetes from
+// their local machine may differ from the one used by pods running in the cluster.
+func (c *clusterRegistrant) respectApiServerOverride(remoteCluster *api.Cluster) error {
```
ilackarms, line 656: nit: "respect" doesn't have a very clear meaning in this context, what about simply `set` instead of `respect`?

ilackarms (approved):

Issue #190, opened 2021-01-07
generate single and multi cluster code for homogenous snapshots


ilackarms:
currently we only generate one or the other based on codegen parameters. to run gloo-mesh discovery as an agent, we will want to generate both

PR #191, opened 2021-01-07
always both single and multicluster snapshot builders

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/190


contrib/codegen/templates/input/input_reconciler.gotmpl
```diff
@@ -168,7 +168,7 @@ func RegisterSingleCluster{{ $snapshotName }}Reconciler(
         predicates ...predicate.Predicate,
 ) (input.InputReconciler, error) {
 
-    base := input.NewInput{{ $snapshotName }}Reconciler(
+    base := input.NewInputReconciler(
```
ilackarms, line 171: this fixes a compile error that occurs if snapshotName is provided

contrib/codegen/templates/input/input_snapshot.gotmpl
```diff
@@ -376,19 +366,28 @@ func (b *multiCluster{{ $snapshotName }}Builder) insert{{ $kindPlural }}FromClus
...
-{{- else }}
-
 // build a snapshot from resources in a single cluster
 type singleCluster{{ $snapshotName }}Builder struct {
     mgr manager.Manager
+    clusterName string
 }
 
-// Produces snapshots of resources across all clusters defined in the ClusterSet
+// Produces snapshots of resources read from the manager for the given cluster
```
harveyxia, line 375:
```suggestion
// Produces snapshots of resources read from the manager for the local cluster
```

ilackarms: Technically you can create one of these from any manager doesn't have to be the local

harveyxia (approved):

PR #196, opened 2021-01-13
Upgrade k8 to v1.19

npolshakova:
Needed for upgrading k8 v1.19 (https://github.com/solo-io/gloo/issues/3582). Bumps k8 to v19 and controller-runtime to v0.7.0. The main change are moving .meta and runtime.Object into the client.Object and some methods now accept a ctx.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/3582


changelog/v0.16.1/update-k8s-19.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NON_USER_FACING
+    description: Update to k8s 0.19.
+    issueLink: https://github.com/solo-io/gloo/issues/3582
```
kdorosh, line 4: nit, `resolvesIssue: false`

changelog/v0.16.2/update-k8s-19.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
```
Sodman, line 1: I don't think we have a hard rule on when to do a minor vs patch on these 0.x repos, but this feels like more of a 0.17.0 I think?

Sodman (commented):
LGTM! @joekelley / @marcogschmidt just wanted to make you both aware of this PR for next time you bump skv2 version in your projects.

changelog/v0.16.2/update-k8s-19.yaml
joekelley: +1, let's mark this as a break as importing projects will probably have to make some changes to integrate this

pkg/events/watcher.go
```diff
@@ -76,7 +77,7 @@ type eventWatcher struct {
 	eventHandler EventHandler
 }
 
-func (w *eventWatcher) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+func (w *eventWatcher) Reconcile(ctx context.Context, request reconcile.Request) (reconcile.Result, error) {
```
EItanya, line 80: Now that `Reconcile` takes a context, can we pass it through to the events?

pkg/controllerutils/upsert.go
```diff
@@ -71,15 +70,12 @@ func transition(existing, desired runtime.Object, transitionFuncs []TransitionFu
...
-	if err != nil {
-		return controllerutil.OperationResultNone, err
-	}
+func UpdateStatus(ctx context.Context, c client.Client, obj client.Object) (controllerutil.OperationResult, error) {
+	key := client.ObjectKeyFromObject(obj)
 
 	existing := obj.DeepCopyObject()
 
-	if err := c.Get(ctx, key, existing); err != nil {
+	if err := c.Get(ctx, key, obj); err != nil {
```
kdorosh, line 78: why are we switching `obj` and `existing` here?

pkg/source/dynamic.go
```diff
@@ -82,7 +82,7 @@ func (s *DynamicSource) Start(h handler.EventHandler, i workqueue.RateLimitingIn
 
 	for _, src := range s.cache {
 
-		if err := src.source.Start(h, i, ps...); err != nil {
+		if err := src.source.Start(s.ctx, h, i, ps...); err != nil {
```
EItanya, line 85: context should not be held, but rather passed into `Start`

pkg/controllerutils/upsert.go
kdorosh: if this is a bug, we should have a test that would have caught this swap

EItanya (changes requested):

pkg/controllerutils/upsert.go
EItanya: Agreed, this requires the same fix as the `Upsert` function above

npolshakova: Yeah that's strange, I have this locally: 

`
	existing := obj.DeepCopyObject().(client.Object)

	if err := c.Get(ctx, key, existing); err != nil {
`

changelog/v0.16.2/update-k8s-19.yaml
EItanya: They will, especially when the `context` related changes are made

pkg/controllerutils/upsert.go
npolshakova: UpdateStatus never got updated- should be fixed now. I can add a unit test for UpdateStatus in the upsert_test

codegen/templates/code/controller/event_handlers.gotmpl
```diff
@@ -88,23 +88,23 @@ type generic{{ $resource.Kind }}Handler struct {
     handler {{ $resource.Kind }}EventHandler
 }
 
-func (h generic{{ $resource.Kind }}Handler) Create(object runtime.Object) error {
+func (h generic{{ $resource.Kind }}Handler) Create(object client.Object) error {
```
ilackarms, line 91: this could be a breaking change for some repos; we should probably mark the changelog as such

pkg/events/watcher.go
ilackarms: do we know what the lifetime of the context is? currently we manage context outside of controller-runtime across our projects. probably would be smart to stage this as a separate change, and be clear about the lifetime (and other properties) of the context that we're using

ilackarms:
overall this lgtm; do you think it would be prohibitive to try importing this version of skv2 into gloo-mesh and/or dev portal to estimate the impact/scope of this change?

EItanya:
> overall this lgtm; do you think it would be prohibitive to try importing this version of skv2 into gloo-mesh and/or dev portal to estimate the impact/scope of this change?

The scope should be very small without the `context` changes. So as you mentioned let's punt on that for a bit.

EItanya:
PR to test against gloo-mesh https://github.com/solo-io/gloo-mesh/pull/1217

EItanya (approved):

Issue #199, opened 2021-01-16
Fix broken controllerutils.ObjectsEqual


ashleywang1:
This is for the case where proto.Message types are passed in by value rather than by pointer.
For example:
```
var a interface{}
a = &fed_types.FailoverSchemeStatus{}
b, ok := a.(proto.Message) // ok == true, this will work
```
```
var a interface{}
a = fed_types.FailoverSchemeStatus{}
b, ok := a.(proto.Message) // ok == false, this will not work
```

The upgrade to using gogo/protobufs resulted in several changes around how protos are considered equal.

Previously, we made a change to fix this:
// DeepEqual should be used in place of reflect.DeepEqual when the type of an object is unknown and may be a proto message.
// see https://github.com/golang/protobuf/issues/1173 for details on why reflect.DeepEqual no longer works for proto messages

The way our reconcilers pass in objects into Upsert may result in these items being passed in by value
```
// in some cases:
type GlooInstance struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Metadata *ObjectMeta               `protobuf:"bytes,1,opt,name=metadata,proto3" json:"metadata,omitempty"`
	Spec     *types.GlooInstanceSpec   `protobuf:"bytes,2,opt,name=spec,proto3" json:"spec,omitempty"`
	Status   *types.GlooInstanceStatus `protobuf:"bytes,3,opt,name=status,proto3" json:"status,omitempty"`
}

// In other cases:
// GlooInstance is the Schema for the glooInstance API
type GlooInstance struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   GlooInstanceSpec   `json:"spec,omitempty"`
	Status GlooInstanceStatus `json:"status,omitempty"`
}
```

Solution: be more flexible in the controllerutils.ObjectsEqual, and be able to deal with both cases.
See: https://github.com/gogo/protobuf/issues/476

ashleywang1:
Additional notes: 

The previous comparison was being done here:
```
equalityutils.DeepEqual:

func DeepEqual(val1, val2 interface{}) bool {
	protoVal1, isProto := val1.(proto.Message)
	if isProto {
		protoVal2, isProto := val2.(proto.Message)
		if !isProto {
			return false // different types
		}
		return proto.Equal(protoVal1, protoVal2)
	}
[...]

PR #200, opened 2021-01-16
make pointer to value if needed

ashleywang1:
This is for the case where proto.Message types are passed in by value rather than by pointer.
For example:
```
var a interface{}
a = &fed_types.FailoverSchemeStatus{}
b, ok := a.(proto.Message) // ok == true, this will work
```
```
var a interface{}
a = fed_types.FailoverSchemeStatus{}
b, ok := a.(proto.Message) // ok == false, this will not work
```

The upgrade to using gogo/protobufs resulted in several changes around how protos are considered equal.

Previously, we made a change to fix this:
// DeepEqual should be used in place of reflect.DeepEqual when the type of an object is unknown and may be a proto message.
// see https://github.com/golang/protobuf/issues/1173 for details on why reflect.DeepEqual no longer works for proto messages

The way our reconcilers pass in objects into Upsert may result in these items being passed in by value
```
// in some cases:
type GlooInstance struct {
	state         protoimpl.MessageState
	sizeCache     protoimpl.SizeCache
	unknownFields protoimpl.UnknownFields

	Metadata *ObjectMeta               `protobuf:"bytes,1,opt,name=metadata,proto3" json:"metadata,omitempty"`
	Spec     *types.GlooInstanceSpec   `protobuf:"bytes,2,opt,name=spec,proto3" json:"spec,omitempty"`
	Status   *types.GlooInstanceStatus `protobuf:"bytes,3,opt,name=status,proto3" json:"status,omitempty"`
}

// In other cases:
// GlooInstance is the Schema for the glooInstance API
type GlooInstance struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec   GlooInstanceSpec   `json:"spec,omitempty"`
	Status GlooInstanceStatus `json:"status,omitempty"`
}
```

Solution: be more flexible in the controllerutils.ObjectsEqual, and be able to deal with both cases.
See: https://github.com/gogo/protobuf/issues/476

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/199

### DIFF
--- a/changelog/v0.16.1/fix-equality.yaml
+++ b/changelog/v0.16.1/fix-equality.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: FIX
+    description: >
+      Fix the case where proto.Message types are passed in by value rather than by pointer.
+    issueLink: https://github.com/solo-io/skv2/issues/199
--- a/pkg/controllerutils/equality_test.go
+++ b/pkg/controllerutils/equality_test.go
@@ -81,6 +81,32 @@ var _ = Describe("ObjectsEqual", func() {
 		equal := ObjectsEqual(obj1, obj2)
 		Expect(equal).To(BeFalse())
 	})
+
+	It("asserts equality on two proto.Message objects even if they are passed in by value", func() {
+		obj1 := &things_test_io_v1.Paint{
+			Spec: things_test_io_v1.PaintSpec{
+				Color: &things_test_io_v1.PaintColor{
+					Hue: "red",
+				},
+				PaintType: &things_test_io_v1.PaintSpec_Oil{
+					Oil: nil,
+				},
+			},
+		}
+		obj1.Spec.ProtoReflect().SetUnknown([]byte(""))
+		obj2 := &things_test_io_v1.Paint{
+			Spec: things_test_io_v1.PaintSpec{
+				Color: &things_test_io_v1.PaintColor{
+					Hue: "red",
+				},
+				PaintType: &things_test_io_v1.PaintSpec_Oil{
+					Oil: nil,
+				},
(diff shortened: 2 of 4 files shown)


changelog/v0.16.1/fix-equality.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: FIX
+    description: >
+      Fix the case where proto.Message types are passed in by value rather than by pointer.
```
EItanya, line 4: How can this happen? golang proto does not support non pointer proto messages

EItanya (changes requested):
lgtm, please add test to verify

pkg/controllerutils/equality_test.go
```diff
@@ -81,6 +81,32 @@ var _ = Describe("ObjectsEqual", func() {
...
+			Spec: things_test_io_v1.PaintSpec{
+				Color: &things_test_io_v1.PaintColor{
+					Hue: "red",
+				},
+				PaintType: &things_test_io_v1.PaintSpec_Oil{
+					Oil: nil,
+				},
+			},
+		}
+		obj1.Spec.ProtoReflect().SetUnknown([]byte(""))
```
ashleywang1, line 96: Previously, when the Spec was being passed in by value, then line
```
protoVal1, isProto := val1.(proto.Message)
```
would return isProto as false, and use reflect.DeepEqual instead of the correct proto.Equal, which allows us to treat empty .status.atomicInfo fields as the same as nil .status.atomicInfo fields.

EItanya (approved):

PR #201, opened 2021-01-18
Remap discovery proto path

Sodman:
[solo-kit](https://github.com/solo-io/solo-kit/blob/bb65ee2700a025bd9faa00555318a1fa4cd192fd/pkg/code-generator/collector/collector.go#L252) remapped the discovery proto. This functionality is needed by some downstream projects depending on skv2 (eg solo-apis), to prevent the discovery.proto namespace conflict errors described in https://github.com/solo-io/gloo/issues/4068.

This change allows adding a `CustomGoArgs` string array, which ultimately gets passed while building protos.

The solo-api's project will add this when generating the skv2 command before Execute()'ing it:
```
CustomGoArgs: []string{"Mgithub.com/solo-io/solo-kit/api/external/envoy/api/v2/discovery.proto=github.com/envoyproxy/go-control-plane/envoy/api/v2"},
```


codegen/collector/compiler.go
```diff
@@ -146,6 +146,7 @@ func (p *protoCompiler) addDescriptorsForFile(addDescriptor func(f DescriptorWit
 
 var defaultGogoArgs = []string{
 	"plugins=grpc",
+	"Mgithub.com/solo-io/solo-kit/api/external/envoy/api/v2/discovery.proto=github.com/envoyproxy/go-control-plane/envoy/api/v2",
```
ashleywang1, line 149: One has the discovery.proto file name and the other doesn't. Is this intentional?

Sodman: I believe so, that's what it's imported as, eg: https://github.com/solo-io/solo-apis/blob/81a712d752f0f38e95b529a864460f8199a2614b/pkg/api/gloo.solo.io/v1/enterprise/ratelimit.pb.go#L17

ashleywang1 (approved):

PR #202, opened 2021-01-20
Common controller entrypoint

ilackarms:
imported from gloo-mesh https://github.com/solo-io/gloo-mesh/blob/main/pkg/common/bootstrap/bootstrap.go

and https://github.com/solo-io/gloo-mesh/blob/main/pkg/common/utils/stats/server.go

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/203


contrib/codegen/funcs/funcs.go
```diff
@@ -58,8 +60,11 @@ func MakeHomogenousSnapshotFuncs(
 	}
 }
 
-func getImportedGroups(selectFromGroups map[string][]model.Group, resourcesToSelect map[schema.GroupVersion][]string) ([]model.Group, map[schema.GroupVersion]importedGroup) {
-	importedGroups := selectResources(selectFromGroups, resourcesToSelect)
+func getImportedGroups(selectFromGroups map[string][]model.Group, resourcesToSelect map[schema.GroupVersion][]string) ([]model.Group, map[schema.GroupVersion]importedGroup, error) {
+	importedGroups, err := selectResources(selectFromGroups, resourcesToSelect)
```
ilackarms, line 64: this change allows us to error when a project's codegen setup is incorrect

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,176 @@
...
+	zaputil "sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
+	// required import to enable kube client-go auth plugins
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
+)
+
+// StartParameters specify paramters for starting a generic controller which may need access to its local cluster as well as remote (multicluster) clients and managers
+type StartParameters struct {
+	Ctx             context.Context
```
EItanya, line 30: nit: can we keep the `context` out of the struct. I personally find it much more complicated to track context when it is not directly passed from function to function.

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,176 @@
...
+type Options struct {
+	// MetricsBindPort is the TCP port that the controller should bind to
+	// for serving prometheus metrics.
+	// It can be set to 0 to disable the metrics serving.
+	MetricsBindPort uint32
+
+	// MasterNamespace if specified restricts the Master manager's cache to watch objects in the desired namespace.
+	// Defaults to all namespaces.
+	//
+	// Note: If a namespace is specified, controllers can still Watch for a cluster-scoped resource (e.g Node).  For namespaced resources the cache will only hold objects from the desired namespace.
```
EItanya, line 56: We should move away from master. maybe `MgmtNamespace`

ilackarms: it's a bit confusing with other uses of management/manager in the code. i don't think there's a clear path forward here, and since this code is already written (just moving repos to make it shared) i'd prefer not to change

EItanya (approved):

PR #204, opened 2021-01-22
add types for generic snapshot

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/213


pkg/resource/snapshot.go
```diff
@@ -0,0 +1,96 @@
...
+		snapshot = Snapshot{}
+	}
+	snapshot.Insert(gvk, obj)
+	cs[cluster] = snapshot
+}
+
+func (cs ClusterSnapshot) Delete(cluster string, gvk schema.GroupVersionKind, id types.NamespacedName) {
+	snapshot, ok := cs[cluster]
+	if !ok {
+		snapshot = Snapshot{}
```
harveyxia, line 73: Can we just return here? Or do we need to initialize a Snapshot entry?

pkg/controllerutils/equality.go
```diff
@@ -107,13 +106,16 @@ func ObjectStatusesEqual(obj1, obj2 runtime.Object) bool {
 		value2 = value2.Elem()
 	}
 
-	status1 := value1.FieldByName("Status")
+	statusField1 := value1.FieldByName("Status")
+	statusField2 := value2.FieldByName("Status")
 
-	if !status1.IsValid() {
-		panic(fmt.Sprintf("cannot compare statuses of object type %T, missing status field", obj1))
+	if !statusField1.IsValid() && !statusField2.IsValid() {
```
harveyxia, line 112: just wondering, which objects don't have statuses?

pkg/bootstrap/bootstrap.go
```diff
@@ -79,9 +85,13 @@ func (opts *Options) AddToFlags(flags *pflag.FlagSet) {
 // Start a controller with the given reconciler. Handles bootstrapping local manager + multicluster watches.
```
harveyxia, line 85:
```suggestion
// Start a controller with the given reconcilers represented by the StartFunc. Handles bootstrapping local manager + multicluster watches.
```

contrib/pkg/input/input_reconciler.go
```diff
@@ -130,7 +130,8 @@ func (r *inputReconciler) processNextWorkItem() bool {
 		requeue bool
 		err     error
 	)
-	if isRemoteCluster {
+	// TODO (ilackarms): this is a workaround for an issue where Relay sets the clusterName field of an object, but the underlying reconciler is a single-cluster reconciler.
```
harveyxia, line 133: Is it no longer true that `clusterName` is only populated for multicluster reconcilers?

Could we modify relay so that this convention continues to hold?

ilackarms: i want to have relay set the cluster name on objects because we are building snapshots of objects that come from remote clusters (e.g. cert issuer) where we rely on cluster name to process the object in the translator

ilackarms: the issue is that we run a single-cluster reconciler to reconcile on local events, but also want to trigger the same reconcile function when a remote object changes (instead of just running a multi cluster reconciler)

pkg/controllerutils/equality.go
ilackarms: configmap, secret, podbouncedirective, etc

pkg/resource/snapshot.go
ilackarms: 😅

harveyxia (approved):

Issue #207, opened 2021-02-01
skv2 protoc casting segfault


ashleywang1:
k8s.io.api.rbac.v1.RoleRef is being cast to a protoc, causing a segfault in the `glooctl cluster register` command.

The following error:
```
panic: field k8s.io.api.rbac.v1.RoleRef.apiGroup has invalid type: got string, want pointer

goroutine 1 [running]:
google.golang.org/protobuf/internal/impl.fieldInfoForScalar(0x4acb8c0, 0xc001345080, 0x41d014a, 0x8, 0x0, 0x0, 0x4acbb00, 0x41ce080, 0x41d0154, 0x34, ...)
        .../go/pkg/mod/google.golang.org/protobuf@v1.25.0/internal/impl/message_reflect_field.go:228 +0xaa7
google.golang.org/protobuf/internal/impl.(*MessageInfo).makeKnownFieldsFunc(0xc000aea500, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xc0017ea210, 0xc0017ea240, 0xc0017ea270, 0xc0017ea2a0)
        .../go/pkg/mod/google.golang.org/protobuf@v1.25.0/internal/impl/message_reflect.go:67 +0xc05
google.golang.org/protobuf/internal/impl.(*MessageInfo).makeReflectFuncs(0xc000aea500, 0x4acbb00, 0x448a220, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xffffffffffffffff, 0xc0017ea210, 0xc0017ea240, 0xc0017ea270, ...)
        .../go/pkg/mod/google.golang.org/protobuf@v1.25.0/internal/impl/message_reflect.go:36 +0x65
google.golang.org/protobuf/internal/impl.(*MessageInfo).initOnce(0xc000aea500)
        .../go/pkg/mod/google.golang.org/protobuf@v1.25.0/internal/impl/message.go:90 +0x325
google.golang.org/protobuf/internal/impl.(*MessageInfo).init(0xc000aea500)
        .../go/pkg/mod/google.golang.org/protobuf@v1.25.0/internal/impl/message.go:72 +0x3f
google.golang.org/protobuf/internal/impl.(*messageReflectWrapper).Range(0xc00106a180, 0xc0017ea1e0)
[...]

PR #208, opened 2021-02-01
Fix skv2 equality

kdorosh:
Fix equality for protos, since we were using the `proto.Equals` on k8s resources that implemented the protov1 interface but should have been using reflection for equality.

The k8s resources don't implement protov2, so use that as our check instead

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/207


ashleywang1:
Tested the fix and `glooctl cluster register` no longer segfaults.

joekelley (approved):

PR #212, opened 2021-02-03
Add gpl license check

saiskee:


ci/oss_compliance/oss_compliance.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/solo-io/go-list-licenses/pkg/license"
+)
+
+func main() {
+	glooPackages := []string{
```
kdorosh, line 11: rename this var?

kdorosh: any reason we hard code this rather than parsing every package?

kdorosh: we should update the logic here to find all the main functions and include those packages. same for gloo and the other repos

there's no chance we remember to update this list if/when we add a new deployable

kdorosh (approved):

Issue #229, opened 2021-04-13
Need UnsortedList on sets


harveyxia:
The current [`sets.List()`](https://github.com/solo-io/skv2/blob/master/contrib/pkg/sets/sets.go#L85) method entails sorting the list of keys. For cases where we need to optimize performance, we should expose an `UnsortedList` on our skv2 set that does not require sorting of keys.

PR #231, opened 2021-04-13
add UpdateStatusImmutable; and change equality utils to only copy when it has to

yuval-k:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/229


changelog/v0.17.15/less-races.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
```
harveyxia, line 1: I think this is the wrong changelog content :)

yuval-k: too much copy paste.. thanks!

harveyxia (commented):
LGTM, I'll let @ilackarms give the approval

ilackarms (approved):

PR #232, opened 2021-04-20
Proto compiler optimizations and metrics

sam-heilbron:
# Work In Progress
I have verified this improves gloo (solo-projects), and have no reason to believe that gloo-mesh will behave differently, but I'd like to confirm that.

# Description

- When compiling protos, use memoization to reduce duplicated work
- Add some basic metrics around duration and frequency of execution, to better understand where bottlenecks are.

# Context

Some recent work in solo-kit (https://github.com/solo-io/solo-kit/pull/415) illustrated that proto compilation (which requires walking a tree of files, accumulating all the imports) takes a long time. This is due to the fact that for each import, we read that file and get those imports. This means, that if we fileA and fileB both import fileC, we will read/parse fileC twice. This grows as the dependency graph gets more complex. 

I added some simple metrics to understand the compilation bottleneck, and ran codegen for solo-projects. During solo-projects codegen, we rely on skv2 twice, first for the gloo fed protos and second for the api server protos. Below are the measurements for our existing implementation of proto compilation compared with the measurements from this change.

### Before (no caching)

Gloo Fed
```
{
    "duration": {
        "@code-generator/proto-compiler": "38.238680152s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 29051
    }
}
```

Api Server
```
{
    "duration": {
        "@code-generator/proto-compiler": "1m2.679936094s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 46697
    }
}
```

### After (with caching)

Gloo Fed
```
{
    "duration": {
        "@code-generator/proto-compiler": "9.828921368s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 145
    }
}
```

Api Server
```
{
    "duration": {
        "@code-generator/proto-compiler": "12.419879844s"
    },
    "frequency": {
        "@code-generator/collector-opened-files": 153
    }
}
```

### DIFF
--- a/codegen/collector/collector.go
+++ b/codegen/collector/collector.go
@@ -1,12 +1,17 @@
 package collector
 
 import (
+	"context"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
+	"time"
+
+	"github.com/solo-io/skv2/codegen/metrics"
 
 	"github.com/rotisserie/eris"
 	"github.com/solo-io/go-utils/log"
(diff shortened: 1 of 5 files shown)


codegen/collector/collector.go
```diff
@@ -131,3 +155,133 @@ func (c *collector) findImportRelativeToRoot(absoluteRoot, importedProtoFile str
...
+	if ok {
+		typedCachedImports := cachedImports.(*fileImports)
+		return typedCachedImports.imports, typedCachedImports.err
+	}
+
+	i.activeRequestsMu.Lock()
+
+	// Ensure that we do not wait forever
+	ctxWithTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
EItanya, line 205: nit: since this called with defer it will only `cancel()` after the select block below, which could potentially block forever

codegen/collector/collector.go
```diff
@@ -131,3 +155,133 @@ func (c *collector) findImportRelativeToRoot(absoluteRoot, importedProtoFile str
...
+	// If there's not a current active request for this file, create one.
+	activeRequest := i.activeRequests[protoFile]
+	if activeRequest == nil {
+		activeRequest = newImportsFetchRequest()
+		i.activeRequests[protoFile] = activeRequest
+
+		// This goroutine has exclusive ownership over the current request.
+		// It releases the resource by nil'ing the importRequest field
+		// once the goroutine is done.
+		go func(requestCtx context.Context) {
```
EItanya, line 216: unused arg

codegen/collector/collector.go
```diff
@@ -131,3 +155,133 @@ func (c *collector) findImportRelativeToRoot(absoluteRoot, importedProtoFile str
 	return possibleImportPaths[0], nil
 
 }
+
+type ImportsFetcher func(file string) ([]string, error)
```
harveyxia, line 159: seems like the `ImportExtractor` logic merits its own file, wdyt?

codegen/collector/collector.go
harveyxia: we shouldn't need to invoke `cancel()` if we're just relying on the timeout

codegen/metrics/metrics.go
```diff
@@ -0,0 +1,80 @@
...
+	"sync"
+	"time"
+)
+
+var (
+	sink *metricSink
+)
+
+func NewSink() {
+	sink = newMetricSink()
```
harveyxia, line 16: nit: the name `MetricsSink` is also Envoy's name for a metrics sink service, which is a different concept. Maybe we can find a different name for this use case?

harveyxia (approved):

Issue #236, opened 2021-04-30
Add first-class annotation support in skv2 deployment helm template generation


ashleywang1:
Needed to add a label to each pod:

https://github.com/solo-io/gloo-mesh/issues/1448

ashleywang1:
Issue with the helm template: https://github.com/solo-io/gloo-mesh/pull/1612#discussion_r644273182

https://github.com/solo-io/skv2/issues/236

PR #237, opened 2021-04-30
 Add additional pod and labels to generated skv2-generated helm chart deployment

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/236


ashleywang1:
> solo-io/gloo-mesh#1450

With the latest change, yes. The helm value would be `[serviceName].extraAnnotations`

ashleywang1:
Tested with https://github.com/solo-io/gloo-mesh/pull/1499 and the helm template rendering worked.

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -12,6 +12,8 @@ Expressions evaluating SKv2 Config use [[ "[[" ]] and [[ "]]" ]]
   [[- $workloadKind = "DaemonSet" ]]
   [[- end -]]
 [[- $operatorVar := (lower_camel $operator.Name) -]]
+[[- $extraPodLabels := $operator.Deployment.Labels -]]
+[[- $extraPodAnnotations := $operator.Deployment.Annotations -]]
```
harveyxia, line 16: Should we put these labels on the deployment as well? I'm not sure about the conventions around Pod labels vs. Deployment labels, usually I see the same set of labels on both

@ilackarms any thoughts?

ashleywang1: I'll add a `extraPodLabels` and `extraDeploymentLabels` field (same for annotations). This allows us to be flexible with which labels/annotations go where.

codegen/model/chart.go
```diff
@@ -51,11 +51,14 @@ type Operator struct {
 // values for Deployment template
 type Deployment struct {
 	// use a DaemonSet instead of a Deployment
-	UseDaemonSet bool                     `json:"-"`
-	Labels       map[string]string        `json:"extraLabels,omitempty"`
-	Annotations  map[string]string        `json:"extraAnnotations,omitempty"`
-	Image        Image                    `json:"image,omitempty"`
-	Resources    *v1.ResourceRequirements `json:"resources,omitempty"`
+	UseDaemonSet    bool                     `json:"-"`
+	PodLabels       map[string]string        `json:"extraPodLabels,omitempty"`
```
harveyxia, line 55: Why name the yaml field differently? Can we instead keep the names of the struct fields consistent with those of the yaml fields?

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -12,8 +12,10 @@ Expressions evaluating SKv2 Config use [[ "[[" ]] and [[ "]]" ]]
   [[- $workloadKind = "DaemonSet" ]]
   [[- end -]]
 [[- $operatorVar := (lower_camel $operator.Name) -]]
-[[- $extraPodLabels := $operator.Deployment.Labels -]]
-[[- $extraPodAnnotations := $operator.Deployment.Annotations -]]
+[[- $extraPodLabels := $operator.Deployment.PodLabels -]]
```
harveyxia, line 15: Same comment here, though more nitpicky, can we use the same name for the template variable as we do for the struct field?

harveyxia (approved):

Issue #246, opened 2021-06-01
Expose arbitrary Kubernetes Deployment and Service spec fields to Helm values


harveyxia:
In addition to the specific Deployment/Service fields currently exposed through Helm values, we should implement a way for user's to specify arbitrary Deployment/Service fields that they wish to control.

A possible implementation approach would be to leverage a json merge semantics in our generated template, where user provided Deployment/Service fields take precedence over the default values shipped with the chart:

```
{{ $defaultSpec := |
spec:
  replicas: {{$.Values.deployment.replicas}}
  #etc...
" }}
{{ $userOverrides := $.Values.deployment.UserOverrides }}
# now the final template
spec: {{ merge $defaultSpec $userOverrides | toYaml }}
```

harveyxia:
@saiskee has implemented something similar in Gloo, [see here(https://github.com/solo-io/gloo/blob/master/install/helm/gloo/templates/1-gloo-deployment.yaml#L196)

PR #248, opened 2021-06-02
Helm deployment and service overrides

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/246


codegen/model/chart.go
```diff
@@ -3,6 +3,8 @@ package model
 import (
 	"fmt"
 
+	appsv1 "k8s.io/api/apps/v1"
+
```
ashleywang1, line 7: nit: no spaces here

codegen/model/chart.go
```diff
@@ -50,24 +55,18 @@ type Operator struct {
...
-	ExtraPodAnnotations        map[string]string        `json:"extraPodAnnotations,omitempty"`
-	Image                      Image                    `json:"image,omitempty"`
-	ExtraDeploymentLabels      map[string]string        `json:"extraDeploymentLabels,omitempty"`
-	ExtraDeploymentAnnotations map[string]string        `json:"extraDeploymentAnnotations,omitempty"`
-	DeploymentImage            Image                    `json:"image,omitempty"`
-	Resources                  *v1.ResourceRequirements `json:"resources,omitempty"`
+	// TODO support use of a DaemonSet instead of a Deployment
+	UseDaemonSet bool                     `json:"-"`
+	Image        Image                    `json:"image,omitempty"`
+	Resources    *v1.ResourceRequirements `json:"resources,omitempty"`
```
ashleywang1, line 61: We will need a Deployment Overrides section too

ilackarms: i don't believe we'll need it here. Deployment overrides are meant to be exposed directly to the user so they can override what we ship in the final chart. the chart itself is something we generate based on the `Operator` type defined in this file. 

So if an skv2 project needs to set some properties on the deployment, it can just configure the `Operator` type being passed to skv2. we can discuss modifying skv2 to expose more aspects of the deployment, but for now i think that's out of scope for this change.

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -97,10 +101,16 @@ spec:
         - name: {{ $[[ $operatorVar ]]Image.pullSecret }}
       {{- end}}
 
+{{- end }} {{/* define "[[ $operator.Name ]].deploymentSpec" */}}
 
----
+{{/* Render [[ $operator.Name ]] deployment template with overrides from values*/}}
+{{- $[[ $operator.Name ]]DeploymentOverrides := dict }}
```
ashleywang1, line 107: If the oprerator.Name is something with a hypen like 'cert-agent', this will resolve to:
```
{{- $cert-agentDeploymentOverrides := dict }}
```
and this error:

Error: parse error at (cert-agent/templates/deployment.yaml:86): bad character U+002D '-'

ashleywang1: Replacing with $[[ $operatorVar ]] will fix it temporarily, but I'm not sure if that $[[ $operatorVar ]] has a guarantee of being in camelcase with no hyphens.

ashleywang1 (changes requested):

codegen/model/chart.go
ashleywang1: Sorry, that was super unclear - we need a way to modify the default Helm Values (e.g. by adding a `sidecar.istio.io/inject : false` annotation)

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -1,60 +1,94 @@
...
+        [[- end ]]
         {{- range $key, $value := $[[ $operatorVar ]]ExtraPodAnnotations }}
         {{ $key }}: {{ $value }}
         {{- end }}
         prometheus.io/path: /metrics
         prometheus.io/port: "9091"
         prometheus.io/scrape: "true"
+        [[- range $key, $value := $customPodAnnotations ]]
+        [[ $key ]]: [[ $value ]]
+        [[- end ]]
```
ashleywang1, line 91: remove this part / it's repeated from above

ashleywang1 (changes requested):

codegen/templates/chart/operator-deployment.yamltmpl
ilackarms: thanks!

codegen/model/chart.go
```diff
@@ -121,12 +124,12 @@ type Values struct {
...
-	ExtraDeploymentLabels      map[string]string `json:"extraDeploymentLabels,omitempty" desc:"Extra labels for the deployment"`
-	ExtraDeploymentAnnotations map[string]string `json:"extraDeploymentAnnotations,omitempty" desc:"Extra annotations for the deployment"`
-	ExtraServiceLabels         map[string]string `json:"extraAnnotations,omitempty" desc:"Extra labels for the service"`
-	ExtraServiceAnnotations    map[string]string `json:"extraAnnotations,omitempty" desc:"Extra annotations for the service"`
+	CustomPodLabels             map[string]string `json:"customPodLabels,omitempty" desc:"Custom labels for the pod"`
+	CustomPodAnnotations        map[string]string `json:"customPodAnnotations,omitempty" desc:"Custom annotations for the pod"`
+	CustomDeploymentLabels      map[string]string `json:"customDeploymentLabels,omitempty" desc:"Custom labels for the deployment"`
+	CustomDeploymentAnnotations map[string]string `json:"customDeploymentAnnotations,omitempty" desc:"Custom annotations for the deployment"`
+	CustomServiceLabels         map[string]string `json:"customAnnotations,omitempty" desc:"Custom labels for the service"`
```
ashleywang1, line 131:
```suggestion
	CustomServiceLabels         map[string]string `json:"customServiceLabels,omitempty" desc:"Custom labels for the service"`
```

ashleywang1 (changes requested):

ashleywang1 (approved):

PR #250, opened 2021-06-16
Add sidecars to operator deployments

ryantking:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/252

### DIFF
--- a/codegen/model/chart.go
+++ b/codegen/model/chart.go
@@ -2,7 +2,6 @@ package model
 
 import (
 	"fmt"
-	"sort"
 
 	"github.com/iancoleman/strcase"
 	"github.com/solo-io/skv2/codegen/doc"
@@ -49,7 +48,7 @@ type Deployment struct {
 	// TODO support use of a DaemonSet instead of a Deployment
 	UseDaemonSet bool
 	Container
-	Sidecars                    map[string]Container
+	Sidecars                    []Sidecar
 	Volumes                     []v1.Volume
 	CustomPodLabels             map[string]string
 	CustomPodAnnotations        map[string]string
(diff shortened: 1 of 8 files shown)


codegen/test/chart/templates/deployment.yaml
```diff
@@ -79,6 +81,7 @@ spec:
           capabilities:
             drop:
             - ALL
+{{/* Sidecars */}}
```
EItanya, line 84: Can we add some sidecars here to prove this out

codegen/model/chart.go
```diff
@@ -35,34 +35,44 @@ type Operator struct {
...
 
 // values for Deployment template
 type Deployment struct {
 	// TODO support use of a DaemonSet instead of a Deployment
-	UseDaemonSet                bool
-	Image                       Image
-	Resources                   *v1.ResourceRequirements
+	UseDaemonSet bool
+	Container
+	Sidecars                    map[string]Container
```
ryantking, line 51: Just want to confirm that since this is going to be a breaking change regardless, we definitely want to have a separate "primary" container and sidecars versus a unified `[]Container`

CC: @ilackarms @EItanya

codegen/model/chart.go
```diff
@@ -35,34 +36,60 @@ type Operator struct {
...
```
ryantking, line 52: Want to confirm that we want to have a separate "primary" container and sidecars vs a single map of all containers. These changes will be breaking regardless.

CC: @ilackarms @EItanya

EItanya (commented):
LGTM, would like a final review from @ilackarms, as well as codegen to be passing

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -123,6 +133,33 @@ spec:
           capabilities:
             drop:
             - ALL
+[[- range $sidecar := $sidecars ]]
+        [[- $sidecarVar := (printf "%s.sidecars.%s" $operatorVar $sidecar.Name) ]]
```
ilackarms, line 137: nit: move this var initialization to line 33 and use consistently

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -123,6 +133,33 @@ spec:
...
+        imagePullPolicy: {{ $[[ $sidecar.Name ]]Image.pullPolicy }}
+        name: [[ $sidecar.Name ]]
+{{- if $[[ $sidecarVar ]].resources }}
+        resources:
+{{ toYaml $[[ $sidecarVar ]].resources | indent 10}}
+{{- else}}
+        resources:
+          requests:
+            cpu: 500m
+            memory: 256Mi
```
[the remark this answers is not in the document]
ryantking: These are the same defaults used in the primary container: https://github.com/solo-io/skv2/pull/250/files#diff-c9c5d9bb4d223c2a38805862fc443457bca61c1de50a149de4a2e7a9ac3466f1L73

ilackarms (approved):

PR #257, opened 2021-06-23
Fix docs generation for extra labels and annotations

ashleywang1:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/246


codegen/model/chart.go
```diff
@@ -124,12 +124,12 @@ type Values struct {
...
 	Env          []v1.EnvVar              `json:"env" desc:"Specify environment variables for the deployment. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#envvarsource-v1-core) for specification details." omitChildren:"true"`
 
-	CustomPodLabels             map[string]string `json:"customPodLabels,omitempty" desc:"Custom labels for the pod"`
-	CustomPodAnnotations        map[string]string `json:"customPodAnnotations,omitempty" desc:"Custom annotations for the pod"`
-	CustomDeploymentLabels      map[string]string `json:"customDeploymentLabels,omitempty" desc:"Custom labels for the deployment"`
-	CustomDeploymentAnnotations map[string]string `json:"customDeploymentAnnotations,omitempty" desc:"Custom annotations for the deployment"`
-	CustomServiceLabels         map[string]string `json:"customServiceLabels,omitempty" desc:"Custom labels for the service"`
-	CustomServiceAnnotations    map[string]string `json:"customServiceAnnotations,omitempty" desc:"Custom annotations for the service"`
+	CustomPodLabels             map[string]string `json:"extraPodLabels,omitempty" desc:"Custom labels for the pod"`
```
EItanya, line 127: Can we change the struct names so they match?

EItanya (changes requested):
Confusing mismatch of `json` name and struct name

codegen/model/chart.go
ashleywang1: Sure! The intention was to differentiate between the labels set by the user (using some helm values.go file) and labels set by the operator (programmatically) - so there'll be another confusing assignment, but it'll be in a different area.

ilackarms (approved):
👍  🎉  thanks for this fix!

PR #276, opened 2021-07-23
add CRD spec hash

yuval-k:


codegen/kuberesource/crd.go
```diff
@@ -15,8 +19,21 @@ import (
...
+func WithoutSpecHash() Option {
+	return func(o *Options) {
+		o.WithoutSpecHash = true
+	}
+}
+
 // Create CRDs for a group
-func CustomResourceDefinitions(group model.Group) (objects []metav1.Object, err error) {
+func CustomResourceDefinitions(group model.Group,
+	opts ...Option) (objects []metav1.Object, err error) {
```
harveyxia, line 36: format nit

```suggestion
func CustomResourceDefinitions(
	group model.Group,
	opts ...Option,
) (objects []metav1.Object, err error) {
```

codegen/kuberesource/crd.go
```diff
@@ -15,8 +19,21 @@ import (
 	"k8s.io/utils/pointer"
 )
 
+type Options struct {
+	WithoutSpecHash bool
+}
+
+type Option func(*Options)
+
+func WithoutSpecHash() Option {
```
harveyxia, line 28: Do we need to represent this option using a function? Could we just pass around a boolean?

codegen/kuberesource/crd.go
```diff
@@ -142,6 +164,16 @@ func CustomResourceDefinition(
...
 		},
 	}
+	if !options.WithoutSpecHash {
+		// hashstructure of the crd spec:
+		specHash, err := hashstructure.Hash(crd.Spec, nil)
+		if err == nil {
+			crd.Annotations = map[string]string{
+				model.CRDSpecHashKey: fmt.Sprintf("%x", specHash),
+			}
+		}
```
harveyxia, line 174: log an error if hashing fails

codegen/kuberesource/crd.go
```diff
@@ -142,6 +164,16 @@ func CustomResourceDefinition(
 			},
 		},
 	}
+	if !options.WithoutSpecHash {
+		// hashstructure of the crd spec:
+		specHash, err := hashstructure.Hash(crd.Spec, nil)
```
harveyxia, line 169: The CRD status can also cause processing errors—can we include the status in the hash as well? Either as a separate annotation or combined with the spec hash

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+type ErrMap map[string]error
+type CrdNeedsUpgrade struct {
+	CRDName string
+}
+
+func (e *CrdNeedsUpgrade) Error() string {
+	return fmt.Sprintf("CRD %s needs to be upgraded", e.CRDName)
+}
+
+type CrdNotFound struct {
```
harveyxia, line 198: Can we simplify by just using an error with a formatted error string?

`eris.Errorf("CRD %s not found. Is it unused?", e.CRDName)`

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+	CRDName string
+}
+
+func (e *CrdNotFound) Error() string {
+	return fmt.Sprintf("CRD %s not found. Is it unused?", e.CRDName)
+}
+
+func DoCrdsNeedUpgrade(newProdCrdInfo model.CRDMetadata, ourCrds []apiextv1beta1.CustomResourceDefinition) ErrMap {
+	newProducutVersion := newProdCrdInfo.Version
```
harveyxia, line 207:
```suggestion
	newProductVersion := newProdCrdInfo.Version
```

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+			needUpgrade, err = DoesCrdNeedUpgrade(newProducutVersion, hash, ourCrd.Annotations)
+		} else {
+			err = &CrdNotFound{CRDName: ourCrd.Name}
+			ret[ourCrd.Name] = err
+		}
+		if err != nil {
+			ret[ourCrd.Name] = err
+		} else if needUpgrade {
+			ret[ourCrd.Name] = &CrdNeedsUpgrade{CRDName: ourCrd.Name}
+		}
```
harveyxia, line 226: tried to take a stab at making the logical flow a bit more readable

```suggestion
		if hash, ok := crdmap[ourCrd.Name]; !ok {
			ret[ourCrd.Name] = &CrdNotFound{CRDName: ourCrd.Name}
			continue
		} else {
			needUpgrade, err := DoesCrdNeedUpgrade(newProducutVersion, hash, ourCrd.Annotations)

			if err != nil {
				ret[ourCrd.Name] = err
			} else if needUpgrade {
				ret[ourCrd.Name] = &CrdNeedsUpgrade{CRDName: ourCrd.Name}
			}
		}
```

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+	crdSpecHash, ok := deployedCrdAnnotations[model.CRDSpecHashKey]
+	if !ok {
+		return false, errors.New(fmt.Sprintf("Cannot determine crd spec hash from CRD annotations: %v", deployedCrdAnnotations))
+	}
+
+	if newCrdHash == crdSpecHash {
+		return false, nil
+	}
+
+	// parse semver of the current product version
```
harveyxia, line 253: move to above L259

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+	newProductVersionSemver, err := semver.NewVersion(newProductVersion)
+	if err != nil {
+		return false, errors.Wrapf(err, "Cannot parse current product version: %s", newProductVersion)
```
harveyxia, line 256:
```suggestion
		return false, errors.Wrapf(err, "Cannot parse new product version: %s", newProductVersion)
```

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+	}
+
+	currentCrdVersionSemvar, err := semver.NewVersion(crdVersion)
```
harveyxia, line 259:
```suggestion
	currentCrdVersionSemver, err := semver.NewVersion(crdVersion)
```

codegen/kuberesource/crd.go
```diff
@@ -153,3 +185,85 @@ func CustomResourceDefinition(
...
+	if err != nil {
+		return false, errors.Wrapf(err, "Cannot parse current crd version: %s", crdVersion)
+	}
+
+	// If the current product version is greater than the crd version, the CRD needs to be upgraded.
```
harveyxia, line 264:
```suggestion
	// If the new product version is greater than the current crd version, the CRD needs to be upgraded.
```

codegen/model/parse/parse.go
```diff
@@ -0,0 +1,69 @@
...
+	"io/fs"
+	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/solo-io/skv2/codegen/model"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+)
+
+func PasreCRDs(f fs.FS) (model.CRDMetadata, error) {
```
harveyxia, line 14:
```suggestion
func ParseCRDs(f fs.FS) (model.CRDMetadata, error) {
```

codegen/model/parse/parse.go
```diff
@@ -0,0 +1,69 @@
...
+	var ret model.CRDMetadata
+	// read all crds from agent-crds and gloo-mesh-crds
```
harveyxia, line 16:
```suggestion
	// read all crds from the specified directory
```

codegen/model/parse/parse.go
```diff
@@ -0,0 +1,69 @@
...
+	// generate a file with all the versions and hashes of the crds
+	// package the generated file into the binary
```
harveyxia, line 18: are these two comments still relevant?

codegen/kuberesource/crd.go
yuval-k: we could but that would break backwards compatability

codegen/kuberesource/crd.go
yuval-k: i believe the schema of the status is included in the crd Spec, so this should cover the status as well

codegen/kuberesource/crd.go
yuval-k: i wanted to a have an error type, so that the check itself can do a type cast. I treat CrdNotFound as a warning and CrdNeedsUpdate as an error, and any other error is ignored and just printed to stdsout.
see the gloo-mesh PR

codegen/kuberesource/crd.go
yuval-k: good point - I think i will actually panic, as this should never happen.

codegen/kuberesource/crd.go
harveyxia: ah my bad—this is the spec of the _CRD definition itself_, not the spec of the CR

codegen/kuberesource/crd.go
```diff
@@ -153,3 +188,90 @@ func CustomResourceDefinition(
 	}
 	return crd
 }
+
+type ErrMap map[string]error
```
ilackarms, line 192: i'm not sure the code here and below belong in `codegen` subpackage. i think any code that is called outside of code generation should live in a `pkg` subdir

codegen/model/chart.go
```diff
@@ -12,6 +12,22 @@ import (
 )
 
 // Chart provides the input data structure for generating Helm charts from the skv2 chart "meta-templates" (templates whose outputs are templates and other files used by generated Helm charts)
+const (
+	CRDVersionKey  = "crd.solo.io/version"
+	CRDSpecHashKey = "crd.solo.io/specHash"
+	CRDMetadataKey = "crd.solo.io/crdMetadata"
+)
+
+type CRDMetadata struct {
```
ilackarms, line 21: nit: keep nested types below top-level types

codegen/kuberesource/crd.go
```diff
@@ -27,7 +46,7 @@ func CustomResourceDefinitions(group model.Group) (objects []metav1.Object, err
 			}
 		}
 
-		objects = append(objects, CustomResourceDefinition(resource, validationSchema))
+		objects = append(objects, CustomResourceDefinition(resource, validationSchema, opts...))
```
ilackarms, line 49: instead of creating/passing the extra `opts...` here, could we not just read `group.SkipSpecHash`?

codegen/kuberesource/crd.go
yuval-k: I've placed it it here mainly because it is related to the logic in this file. what package do you think it should go in?

codegen/model/chart.go
yuval-k: CRDMetadata is not a nested type?

codegen/kuberesource/crd.go
ilackarms: id say either create a new package like `pkg/crdutils` or use existing like `pkg/controllerutils`

codegen/model/chart.go
ilackarms: i see. i was confused because `chart.go` contains the types that are used to provide a model for how the chart should be generated. do you think it makes sense to expose these structs from the same package? also it seems like this is code could be used outside of code generation

ilackarms (approved):

PR #283, opened 2021-08-12
chart gen: add flag to enable/disable operators

ilackarms:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/284


codegen/test/chart/templates/deployment.yaml
```diff
@@ -120,39 +122,45 @@ spec:
       imagePullSecrets:
         - name: {{ $painterImage.pullSecret }}
       {{- end}}
-
+{{- end }}
```
ilackarms, line 125: i think we can remove the enabled check from the definition since we do the check around where we call the definiton, wdyt?

npolshakova (approved):

PR #285, opened 2021-08-17
Verify resource exists before sync

npolshakova:
When syncing the output snapshot to local cluster storage only write resources that exist. 
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/987

### DIFF
--- a/contrib/pkg/output/snapshot.go
+++ b/contrib/pkg/output/snapshot.go
@@ -249,7 +249,6 @@ func (s Snapshot) SyncMultiCluster(ctx context.Context, mcClient multicluster.Cl
 		// if the cluster is not available, we will not error (simply skip writing the resources here)
 		for _, cluster := range s.Clusters {
 			listForCluster := listsByCluster[cluster]
-			var presentResources []ezkube.Object
 
 			cli, err := mcClient.Cluster(cluster)
 			if err != nil {
@@ -260,22 +259,8 @@ func (s Snapshot) SyncMultiCluster(ctx context.Context, mcClient multicluster.Cl
 				continue
 			}
 
-			for _, resource := range listForCluster {
-				if opts.Verifier != nil {
-					gvk := resource.GetObjectKind().GroupVersionKind()
-					resourceRegistered, _ := opts.Verifier.VerifyServerResource(
-						cluster,
-						gvk,
-					)
-
-					if resourceRegistered {
-						presentResources = append(presentResources, resource)
-					}
-				}
-			}
-
 			resourcesForCluster := ResourceList{
-				Resources: presentResources,
+				Resources: listForCluster,
 				ListFunc:  list.ListFunc,
 				GVK:       list.GVK,
 			}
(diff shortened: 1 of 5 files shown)


contrib/pkg/output/snapshot.go
```diff
@@ -207,53 +208,91 @@ type Snapshot struct {
 	ListsToSync []ResourceList
 }
 
+// Options for writing resources of a given type
+type OutputOpts struct {
```
ashleywang1, line 212: nice!

contrib/pkg/output/snapshot.go
```diff
@@ -207,53 +208,91 @@ type Snapshot struct {
...
-					errHandler.HandleWriteError(object, err)
+					opts.ErrHandler.HandleWriteError(object, err)
 				}
 				continue
 			}
 
+			for _, resource := range listForCluster {
+				if opts.Verifier != nil {
+					gvk := resource.GetObjectKind().GroupVersionKind()
+					resourceRegistered, _ := opts.Verifier.VerifyServerResource(
```
ashleywang1, line 278: Can we either debug log the error here, or add a comment on why it's ok to ignore it?

contrib/pkg/output/snapshot.go
```diff
@@ -207,53 +208,91 @@ type Snapshot struct {
...
+func (s Snapshot) SyncLocalCluster(ctx context.Context, cli client.Client, opts OutputOpts) {
 
+	var presentResources []ezkube.Object
 	for _, list := range s.ListsToSync {
 		listForLocalCluster := list.SplitByClusterName()[multicluster.LocalCluster]
 
+		for _, resource := range listForLocalCluster {
+			if opts.Verifier != nil {
+				gvk := resource.GetObjectKind().GroupVersionKind()
+				resourceRegistered, _ := opts.Verifier.VerifyServerResource(
```
ilackarms, line 232: we should check the error and report it to the err handler (and `continue`)

contrib/pkg/output/snapshot.go
```diff
@@ -207,53 +208,91 @@ type Snapshot struct {
...
 // only writes resources intended for the local cluster (with ClusterName == "")
 // Note that Syncing snapshots in this way adds the label
-func (s Snapshot) SyncLocalCluster(ctx context.Context, cli client.Client, errHandler ErrorHandler) {
+func (s Snapshot) SyncLocalCluster(ctx context.Context, cli client.Client, opts OutputOpts) {
 
+	var presentResources []ezkube.Object
 	for _, list := range s.ListsToSync {
 		listForLocalCluster := list.SplitByClusterName()[multicluster.LocalCluster]
 
+		for _, resource := range listForLocalCluster {
```
ilackarms, line 229: i don't think we need to iterate through each resource in each list; just check the gvk for each list

contrib/pkg/output/snapshot.go
```diff
@@ -207,53 +208,91 @@ type Snapshot struct {
...
+			if opts.Verifier != nil {
+				gvk := resource.GetObjectKind().GroupVersionKind()
```
ilackarms, line 231: we don't set these when we translate out objects, we may need to use `apiutil.GVKForObject` here. another option is to just pass the gvks in from generated code

ilackarms (approved):

PR #290, opened 2021-08-31
Limit concurrent protoc instances

ryantking:


ilackarms (dismissed):

ilackarms (approved):

codegen/collector/compiler.go
```diff
@@ -77,7 +78,19 @@ func (p *protoCompiler) CompileDescriptorsFromRoot(root string, skipDirs []strin
...
+		group            errgroup.Group
+		sem              chan struct{}
+		limitConcurrency bool
+	)
+	if s := os.Getenv("MAX_CONCURRENT_PROTOCS"); s != "" {
+		maxProtocs, err := strconv.Atoi(s)
+		if err != nil {
+			return nil, eris.Wrapf(err, "invalid value for MAX_CONCURRENT_PROTOCS: %s", s)
+		}
+		sem = make(chan struct{}, maxProtocs)
```
kfelter, line 91: Does this chan need to be closed at some point?

ryantking: Original PR author here: No, closing go channels is only required when the business logic requires that you indicate that all data has been sent through the channel. Since channels are not backed by a file descriptor like a network connection or file on disk, closing them is purely semantic. Once the open channel goes out of scope it will be garbage collected like any other variable.

PR #293, opened 2021-09-08
AddToScheme from SchemeBuilder

jahvon:
This change is needed because the upstream Istio `IstioOperator` CR does not define `AddToScheme` in the import package that we use. We can also get their through `SchemeBuilder` so this change impacts all generated clients


EItanya:
LGTM, @ilackarms wdyt?

EItanya (approved):

PR #299, opened 2021-09-23
v1-crd

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/300


pkg/multicluster/internal/k8s/apiextensions.k8s.io/v1/clients.go
```diff
@@ -0,0 +1,216 @@
+// Code generated by skv2. DO NOT EDIT.
```
jahvon, line 1: Should the v1beta1 counterpart to all of these generated files be removed?

EItanya: it doesn't really matter, it's just clients, we can have both

jahvon (approved):

PR #303, opened 2021-10-11
Various fixes

EItanya:


pkg/bootstrap/bootstrap.go
```diff
@@ -65,8 +65,9 @@ type Options struct {
 	// enables verbose mode
 	VerboseMode bool
 
-	// enables json logger (instead of table logger)
-	JSONLogger bool
```
ashleywang1, line 69: The rest LGTM - but maybe update the changelog to mention that this option flag only affects the StartMulti function?

PR #307, opened 2021-10-14
Skip files

jahvon:


codegen/model/parse/parse.go
```diff
@@ -15,7 +15,7 @@ import (
  * This is used by downstream projects to collect the spec hashes of all CRDs, and the version of the product.
  * the resulting CRDMetadata will most likely end up on the downstream project's deployment annotations.
  */
-func ParseCRDs(f fs.FS) (crdutils.CRDMetadata, error) {
+func ParseCRDs(f fs.FS, ignoredFiles map[string]struct{}) (crdutils.CRDMetadata, error) {
```
EItanya, line 18: Super nit: we typically use lists for this sort of filtering

jahvon: I figured a map would be nice here for faster look-up but I can just change this to a list

Issue #311, opened 2021-11-18
`x-kubernetes-preserve-unknown-fields` is not included in the generated CRD validation schema for unstructured maps


jahvon:
When the `disable_openapi_validation` is set on a map value, the CRD validation schema does not include the expected `x-kubernetes-preserve-unknown-fields` value.

PR #312, opened 2021-11-18
Fix unstructured map schema gen

jahvon:
When the `disable_openapi_validation` is set on a map value, the CRD validation schema does not include the expected `x-kubernetes-preserve-unknown-fields` value. This PR fixes that by removing "value" from the field path so that Cue can correctly identify the field.

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/311


codegen/proto/options.go
```diff
@@ -71,7 +80,6 @@ func (o Options) getUnstructuredFields(protoPkg string, rootMessage []string) ([
 			// the field is a primitive type
 			continue
 		case descriptor.FieldDescriptorProto_TYPE_MESSAGE:
-			// TODO: verify this works with map types
```
ilackarms, line 74: 👍

Issue #317, opened 2022-01-24
Incorrect group in generated snapshot


jenshu:
Generating a snapshot with resources from multiple groups results in all resources incorrectly being assigned the same group in the generated snapshot. This regression started in skv2 v0.22.1

PR #318, opened 2022-01-24
Fix resource group in generated snapshot

jenshu:
Fix a bug that occurred when generating a snapshot with resources from multiple groups. The snapshot resources now have the correct groups attached to them.
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/317

### DIFF
--- a/codegen/model/resource.go
+++ b/codegen/model/resource.go
@@ -134,13 +134,13 @@ func (g *Group) InitDescriptors(descriptors []*collector.DescriptorWithPath) {
 // ensures the resources point to this group
 func (g *Group) Init() {
 	for i, resource := range g.Resources {
-		resource.Group = *g
+		resource.Group = g
 		g.Resources[i] = resource
 	}
 }
 
 type Resource struct {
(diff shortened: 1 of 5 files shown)


codegen/model/resource.go
```diff
@@ -134,13 +134,13 @@ func (g *Group) InitDescriptors(descriptors []*collector.DescriptorWithPath) {
 // ensures the resources point to this group
 func (g *Group) Init() {
 	for i, resource := range g.Resources {
-		resource.Group = g
+		resource.Group = *g
```
ilackarms, line 137: i'm not 100% sure what this might affect in case we need this somewhere else, but there is a better way to solve this bug when iterating over a slice of pointers:

```
for _, ptr := range ptrValues {
    ptr := ptr // shadow the variable ensures the correct value will be used
    // act on ptr here
}
```

jenshu: ah you're right, i had been trying to do that but didn't find the right spot. fixed now

ilackarms (approved):

PR #339, opened 2022-05-13
Fix codegen where spec or status fields are types defined in another package

birkland:
The following is an snippet of a Group that skv2 would have trouble rendering:
```
        return []model.Group{
                {
                        GroupVersion: schema.GroupVersion{
                                Group:   "extauth.solo.io",
                                Version: "v1",
                        },
                        Module:  "github.com/solo-io/ext-auth-service",
                        ApiRoot: "pkg/api",
                        Resources: []model.Resource{
                                {
                                        Kind: "AuthConfig",
                                        Spec: model.Field{
                                                Type: model.Type{
                                                        Name:         "AuthConfigSpec",
                                                        ProtoPackage: "enterprise.gloo.solo.io",
                                                        GoPackage:    "github.com/solo-io/solo-apis/api/gloo/enterprise.gloo/v1",
                                                },
                                        },
                                        Status: &model.Field{
                                                Type: model.Type{
                                                        Name: "AuthConfigStatus",
                                                },
                                        },
                                },
                        },
 ```
 In this example, the AuthConfig.Spec is actually a type pulled in from another package (github.com/solo-io/solo-apis/api/gloo/enterprise.gloo/v1), whereas Authconfig.Status is a field that is generated from a .proto definition in the current package.
 
Typical problems include :
* rendering "dot" imports that can result in name conflicts of types
* Attempting to implement methods for types defined in other packages

This PR smooths out the rough edges for such constructs and results in syntactically correct code.

### CODE REVIEW

File: codegen/render/manifests_renderer.go [hunk 1]
```diff
@@ -329,3 +331,11 @@ func removeProtoAnyValidation(d map[string]interface{}) {
 		values["x-kubernetes-preserve-unknown-fields"] = true
 	}
 }
+
+func ifDefined(val, defaultVal string) string {
```
ilackarms, line 335: nit: how about `defaultValue`

codegen/render/proto_types_renderer.go
```diff
@@ -199,3 +204,18 @@ func uniqueGoImportPathsForGroup(grp Group) []string {
 	sort.Strings(result)
 	return result
 }
+
+// We can only define methods for types defined in the same package as the root,
+// so elide resources that neither hace a spec nor a status in the root pkg (a rare edge case)
```
marcogschmidt, line 209: Super super nit:
```suggestion
// so elide resources that neither have a spec nor a status in the root pkg (a rare edge case)
```

codegen/util/packages.go
```diff
@@ -42,3 +44,9 @@ func GeneratedGoPackage(grp model.Group) string {
 
 	return s
 }
+
+// Generate a package alias, eg 'import alias github.com/my/package'
+// TODO:  Do something prettier if this works
```
marcogschmidt, line 49: Is this TODO to be addressed in this PR? If not, maybe we should add an issue for this and reference it here.

codegen/templates/code/types/json.gen.gotmpl
```diff
@@ -24,6 +24,7 @@ var (
 
 {{- range $resource := $.Resources }}
 
+{{- if or (eq $resource.Spec.Type.GoPackage "") (eq $resource.Spec.Type.GoPackage $.RootGoPackage)}}
```
marcogschmidt, line 27: Nit: I think you could simplify this to
```suggestion
{{- if or $resource.Spec.Type.GoPackage (eq $resource.Spec.Type.GoPackage $.RootGoPackage)}}
```

marcogschmidt (commented):
Just one question but otherwise LGTM

codegen/util/packages.go
birkland: No, it's not.  Ultimately, it's just a style thing.  I think creating an issue makes sense, then we can decide if it's worth it for increased readability of the types.go file

marcogschmidt (dismissed):

marcogschmidt (approved):

PR #340, opened 2022-05-16
fix(watcher.go): prevent compute-halting when observing a cluster wit…

gunnar-solo:
…hout `gloo`

related https://github.com/solo-io/gloo/issues/5832


pkg/multicluster/watch/watcher.go
```diff
@@ -94,7 +94,14 @@ func (c *clusterWatcher) startManager(clusterName string, mgr manager.Manager) {
...
 		err := mgr.Start(ctx)
 		if err != nil {
-			contextutils.LoggerFrom(ctx).DPanicf("manager start failed for cluster %v: %v", clusterName, err)
+			/* err has been downgraded from DPanicf --> Errorf.  This case is encountered consistently in `gloo` when a client
+			   attempts to run `glooctl cluster register` on a cluster that _does not contain_ a `gloo` install.  Previously, this would put
+			   the `gloo-fed` pod into a failure state, which prevented...
+			   	* new clusters from being registered
+				* old cluster from successfully being registered
+			*/
+			contextutils.LoggerFrom(ctx).Errorf("manager start failed for cluster %v: %v.  This is likely because a `gloo` instance is not installed to %v.  Recommendation:  run `glooctl cluster deregister --cluster-name=kind-remote` to deregisted the failed `remote`, install `gloo`, then attempt again.", clusterName, err, clusterName)
```
EItanya, line 103: The only issue I see here is that there’s no recovery or retry here, so the pod could potentially be in a bad state in perpetuity

PR #342, opened 2022-06-06
Allow users to set container security context

jmhbh:


codegen/model/values/helm_chart_values.go
```diff
@@ -37,9 +37,10 @@ type UserValues struct {
...
+	Image           Image                    `json:"image" desc:"Specify the container image"`
+	Env             []v1.EnvVar              `json:"env" desc:"Specify environment variables for the container. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#envvarsource-v1-core) for specification details." omitChildren:"true"`
+	Resources       *v1.ResourceRequirements `json:"resources,omitempty" desc:"Specify container resource requirements. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#resourcerequirements-v1-core) for specification details."`
+	SecurityContext *v1.SecurityContext      `json:"securityContext,omitempty" desc:"Specify container security context. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#securitycontext-v1-core) for specification details."`
```
arianaw66, line 43: Should we extend the description to give a summary of the default if the field is not set?

arianaw66 (commented):
The implementation looks good to me, though I'd prefer is someone more familiar with skv2/helm gave final approval.

However, I don't think I follow why we decided to allow for `--set glooMeshRedis.server.securityContext=false`. It seems like `--set glooMeshRedis.server.securityContext={}` would have the same effect if I understand correctly? And the mixed type seems confusing, the logic in the code a tad more complex, and I think we've been trying to avoid multiple ways of doing the same thing.

Thoughts? Or explanations if I missed it/misunderstood?

jmhbh:
> Gotcha. I had misunderstood. Would `--set glooMeshRedis.server.securityContext=null` work? I believe Marco maybe wanted to avoid that one, but just want to confirm.

There's a complication with using `null` because by default `securityContext` is not set and it defaults to `nil` which we use to then render the default `securityContext`. So if we modify the line [here](https://github.com/solo-io/skv2/blob/468ca1bf47fc4a1c53045c8dd8d7f74e02889790/codegen/templates/chart/operator-deployment.yamltmpl#L112) to `{{- else if eq nil [[ $containerVar ]].securityContext }}` we won't be rendering the default `securityContext` when we should be.

jmhbh:
> > `eshRedis.server.securityContext={}` would have the same effect if I understand correctly? And the mixed type seems confusing, the logic in the code a tad more complex, and I think we've been trying to avoid multiple ways of doing the same thing.
> > Thoughts? Or explanations if I missed it/misunderstood?

To your point of the mixed type, I'm not a big fan of that as well. However, without adding that we wouldn't be able to support the user omitting the `securityContext` via the helm client using `--set glooMeshRedis.server.securityContext=false` they would have to use the `-f` option as I have described in the description of this [PR](https://github.com/solo-io/gloo-mesh-enterprise/pull/4083)

arianaw66:
Okay, makes sense. Thanks for explaining all that!

codegen/model/values/helm_chart_values.go
jmhbh: Oh ok got it. So I think we can actually omit it here, but I will update the helm value references in this [PR](https://github.com/solo-io/gloo-mesh-enterprise/pull/4083) to contain the default values. Those will be user facing on our docs [here](https://docs.solo.io/gloo-mesh-enterprise/latest/reference/helm/gloo_mesh_enterprise/latest/gloo_mesh_enterprise_helm_values_reference/) would that work?

codegen/test/chart-no-desc/templates/deployment.yaml
```diff
@@ -100,7 +111,17 @@ spec:
...
 {{- end}}
+        {{/*
+          Render securityContext configs if it is set or an empty map.
+          Otherwise if it is false render an empty map. If securityContext is not set render default securityContext.
+        */}}
         securityContext:
+{{- if or ($palette.securityContext) (eq "map[]" (printf "%v" $palette.securityContext)) }}
+{{ toYaml $palette.securityContext | indent 10}}
+{{/* Because securityContext is nil by default we can only perform following conversion if it is a boolean. Skip conditional otherwise. */}}
+{{- else if eq (ternary $palette.securityContext true (eq "bool" (printf "%T" $palette.securityContext))) false }}
```
jmhbh, line 122: this got more complicated and I don't love it but I needed to add this as the GME helm tests were running into the following error:

```
<string>: Error: template: gloo-mesh-enterprise/templates/deployment.yaml:111:3: executing "gloo-mesh-enterprise/templates/deployment.yaml" at <include "skv2.utils.merge" (list . $glooMeshMgmtServerDeploymentOverrides "gloo-mesh-mgmt-server.deploymentSpec")>: error calling include: template: gloo-mesh-enterprise/templates/_helpers.tpl:47:22: executing "skv2.utils.merge" at <include (index . 2) $top>: error calling include: template: gloo-mesh-enterprise/templates/deployment.yaml:85:12: executing "gloo-mesh-mgmt-server.deploymentSpec" at <eq false $glooMeshMgmtServer.securityContext>: error calling eq: invalid type for comparison```

**Comment by ashleywang1** [comment]
This approach looks good, but I'm seeing errors like:
```
Error: template: gloo-mesh-enterprise/templates/deployment.yaml:111:3: executing "gloo-mesh-enterprise/templates/deployment.yaml" at <include "skv2.utils.merge" (list . $glooMeshMgmtServerDeploymentOverrides "gloo-mesh-mgmt-server.deploymentSpec")>: error calling include: template: gloo-mesh-enterprise/templates/_helpers.tpl:47:22: executing "skv2.utils.merge" at <include (index . 2) $top>: error calling include: template: gloo-mesh-enterprise/templates/deployment.yaml:85:12: executing "gloo-mesh-mgmt-server.deploymentSpec" at <eq false $glooMeshMgmtServer.securityContext>: error calling eq: invalid type for comparison
make: *** [Makefile:342: setup-test-clusters] Error 1
```
on the GME PR https://github.com/solo-io/gloo-mesh-enterprise/pull/4083

This PR should stay open until the GME PR passes, and then I can approve both.

File: codegen/templates/chart/operator-deployment.yamltmpl [hunk 3]
```diff
@@ -102,7 +102,17 @@ spec:
             cpu: 500m
             memory: 256Mi
 {{- end}}
+        {{/*
+          Render securityContext configs if it is set or an empty map.
+          Otherwise if it is false render an empty map. If securityContext is not set render default securityContext.
```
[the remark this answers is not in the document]
ashleywang1: Here and below

ashleywang1 (approved):

codegen/templates/chart/operator-deployment.yamltmpl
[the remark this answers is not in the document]
jmhbh: addressed these comments in the new [PR](https://github.com/solo-io/skv2/pull/346)

PR #346, opened 2022-06-13
Allow users to set container security context (new)

jmhbh:
**Overview**
* Original [PR](https://github.com/solo-io/skv2/pull/342) was merged accidentally this is the new PR
* In this [issue](https://github.com/solo-io/gloo-mesh-enterprise/issues/4000) the ask is to allow the user to omit container `securityContext` using helm's `--set` options when running helm install for the GME enterprise chart specifically to omit the container `securityContext` for the `glooMeshRedis` container.
* Because the GME helm charts are generated via SKV2. In order to address this, the change in this PR allows the container `securityContext` to be overridden and modifies the  `codegen/templates/chart/operator-deployment.yamltmpl` file to conditionally render the override `securityContext` configurations if it exists or is an empty map. In order to support the user needing to omit the `securityContext` via the helm cli we will also check to see if `securityContext` is false and render an empty map if so (this will support the user appending `--set glooMeshRedis.securityContext=false` to their helm install command). Otherwise the default container security context config is rendered. 
* For more context into why this approach was taken please refer to my comment on the issue [here](https://github.com/solo-io/gloo-mesh-enterprise/issues/4000#issuecomment-1149955422)


jmhbh:
> This approach looks good, but I'm seeing errors like:
> 
> ```
> Error: template: gloo-mesh-enterprise/templates/deployment.yaml:111:3: executing "gloo-mesh-enterprise/templates/deployment.yaml" at <include "skv2.utils.merge" (list . $glooMeshMgmtServerDeploymentOverrides "gloo-mesh-mgmt-server.deploymentSpec")>: error calling include: template: gloo-mesh-enterprise/templates/_helpers.tpl:47:22: executing "skv2.utils.merge" at <include (index . 2) $top>: error calling include: template: gloo-mesh-enterprise/templates/deployment.yaml:85:12: executing "gloo-mesh-mgmt-server.deploymentSpec" at <eq false $glooMeshMgmtServer.securityContext>: error calling eq: invalid type for comparison
> make: *** [Makefile:342: setup-test-clusters] Error 1
> ```
> 
> on the GME PR [solo-io/gloo-mesh-enterprise#4083](https://github.com/solo-io/gloo-mesh-enterprise/pull/4083)
> 
> This PR should stay open until the GME PR passes, and then I can approve both.

Thanks for your comments @ashleywang1! I fixed this issue as can be seen in the helm test workflow in my GME [PR](https://github.com/solo-io/gloo-mesh-enterprise/pull/4083). Also addressed your other comments.

codegen/model/values/helm_chart_values.go
```diff
@@ -37,9 +37,10 @@ type UserValues struct {
...
+	Image           Image                    `json:"image" desc:"Specify the container image"`
+	Env             []v1.EnvVar              `json:"env" desc:"Specify environment variables for the container. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#envvarsource-v1-core) for specification details." omitChildren:"true"`
+	Resources       *v1.ResourceRequirements `json:"resources,omitempty" desc:"Specify container resource requirements. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#resourcerequirements-v1-core) for specification details."`
+	SecurityContext *v1.SecurityContext      `json:"securityContext,omitempty" desc:"Specify container security context. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#securitycontext-v1-core) for specification details. Can be set to false to omit rendering of securityContext for the container."`
```
ashleywang1, line 43:
```suggestion
	SecurityContext *v1.SecurityContext      `json:"securityContext,omitempty" desc:"Specify container security context. Set to 'false' to omit the security context entirely. See the [Kubernetes documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.20/#securitycontext-v1-core) for specification details. Can be set to false to omit rendering of securityContext for the container."`
```

ashleywang1: Here and below

ashleywang1 (changes requested):

codegen/model/values/helm_chart_values.go
jmhbh: fixed

ashleywang1 (approved):

Issue #347, opened 2022-06-17
Improve snapshot endpoints & UI


josh-pritchard:
- Add new params to `snapshots/output` & `snapshots/input`
  - format
  - clusters
  - namespaces
  - resourceTypes
- Improve UI to utilize new options

PR #348, opened 2022-06-17
Improve snapshot endpoints & UI

josh-pritchard:
Endpoints now support things like this `http://localhost:9091/snapshots/input?format=yaml&clusters=cluster-1::cluster-2&namespaces=bookinfo`

New URL params:
- format
- clusters
- namespaces
- resourceTypes
Improved UI to utilize new options & added direct download button.
image

### Fallback UI
image

Tested various filtering options extensively.
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/347


pkg/stats/io_snapshots.go
```diff
@@ -37,25 +42,78 @@ func (h *snapshotHistory) SetInput(id string, latestInput json.Marshaler) {
...
+			}
 		}
-		result[id] = genericMap
 	}
-	return json.Marshal(result)
+
+	// If filters enabled only return translator section
+	if clusters != nil || namespaces != nil || resourceTypes != nil {
+		return formatMap(format, translator)
+	} else {
```
jehawley, line 100: Should this just short-circuit all of the filtering logic and return early if it can?

pkg/stats/io_snapshots.go
```diff
@@ -37,25 +42,78 @@ func (h *snapshotHistory) SetInput(id string, latestInput json.Marshaler) {
...
+				delete(translator, k)
+			}
 		}
-		genericMap := map[string]interface{}{}
-		if err := json.Unmarshal(jsn, &genericMap); err != nil {
-			return nil, err
+	}
+
+	// Filter cluster and namespaces
+	if clusters != nil || namespaces != nil {
```
jehawley, line 67: Are you sure you want the behavior to be different in the cases where `clusters == nil` and `clusters != nil && len(clusters) == 0`? It seems rare that you would actually want the behavior where you return a result with nothing selected.

pkg/stats/index.go
```diff
@@ -19,24 +118,111 @@ var indexTmpl = template.Must(template.New("index").Parse(`<!DOCTYPE html><html>
...
+{{range .Profiles}}
 <h2><a href={{.Href}}>{{.Name}}</a></h2>
 <p>
 {{.Desc}}
 </p>
 {{end}}
+<br>
+<br>
+<h2>Snapshot Format & Filters</h2>
+<p>All options apply to both snapshots. Filtering is inclusive and filters strictly by input resources.</p>
```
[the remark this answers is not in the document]
josh-pritchard: Yeah that isn't very clear. Will revise

changelog/v0.22.19/snapshot_ui_and_endpoint_improvements.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
+  - type: NEW_FEATURE
+    description: Improve snapshot endpoints & UI.
```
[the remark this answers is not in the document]
josh-pritchard: Will add to this

pkg/stats/index.go
```diff
@@ -19,24 +118,111 @@ var indexTmpl = template.Must(template.New("index").Parse(`<!DOCTYPE html><html>
...
+		return
+	}
+	translator := input["translator"].(map[string]interface{})
+	resourceTypes := make([]string, 0, len(translator))
+	for k := range translator {
+		resourceTypes = append(resourceTypes, k)
+	}
+
+	for k, resources := range translator {
+		for _, resource := range resources.([]interface{}) {
```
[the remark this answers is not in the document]
josh-pritchard: They are. This is a symptom of working with generic types for this data

pkg/stats/index.go
```diff
@@ -51,5 +237,17 @@ func Index(w http.ResponseWriter, r *http.Request) {
 		return profiles[i].Name < profiles[j].Name
 	})
 
-	indexTmpl.Execute(w, profiles)
+	data := pageData{
+		Profiles:      profiles,
+		Formats:       []string{"", "json", "yaml"},
+		ResourceTypes: resourceTypes,
+		Clusters:      clusterSlice,
+		Namespaces:    namespaceSlice,
```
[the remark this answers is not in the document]
josh-pritchard: I think this is okay for now. It would have to be an extremely large deployment to cause issues with the browser and at that point I am much more concerned about the size of the in-memory snapshots we are maintaining.

pkg/stats/io_snapshots.go
```diff
@@ -37,25 +42,78 @@ func (h *snapshotHistory) SetInput(id string, latestInput json.Marshaler) {
...
 	defer h.lock.RUnlock()
 	if h.latestInput == nil {
 		return []byte{}, nil
 	}
-	result := map[string]interface{}{}
-	for id, obj := range h.latestInput {
-		jsn, err := obj.MarshalJSON()
-		if err != nil {
-			return nil, err
+	genericMaps, err := getGenericMaps(h.latestInput)
```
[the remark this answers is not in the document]
josh-pritchard: Discussed this over phone with John but in general I went with this approach because we are already taking a full copy of the snapshot initially so we don't have to lock it long. Working with that copy to create the result I believe is more memory efficient than working with another new object for the result.

pkg/stats/index.go
```diff
@@ -19,24 +119,111 @@ var indexTmpl = template.Must(template.New("index").Parse(`<!DOCTYPE html><html>
...
+	resourceTypes := make([]string, 0, len(translator))
+	for id := range translator {
+		resourceTypes = append(resourceTypes, id)
+	}
+
+	for resourceType, resources := range translator {
+		for _, resource := range resources.([]interface{}) {
+			resource := resource.(map[string]interface{})
+			metadata := resource["metadata"].(map[string]interface{})
+			cluster, ok := metadata["clusterName"].(string)
```
Sodman, line 202: Is this populated for input data? I had a feeling we only populated this for the output snapshot. K8s has announced they'll be deprecated the clusterName field soon, so it's not widely used atm, even though we use it in our output snapshots.

pkg/stats/io_snapshots.go
```diff
@@ -37,25 +77,81 @@ func (h *snapshotHistory) SetInput(id string, latestInput json.Marshaler) {
...
+	if err != nil {
+		return nil, err
+	}
+	return genericMaps, nil
+}
+
+// DEPRECIATED - Use GetFilteredInput or GetInputCopy instead
+func (h *snapshotHistory) GetInput() ([]byte, error) {
```
Sodman, line 94: Rather than deprecating, could we just have this wrap GetFilteredInput with no filters?

pkg/stats/io_snapshots.go
```diff
@@ -64,35 +160,135 @@ func (h *snapshotHistory) SetOutput(id string, latestOutput json.Marshaler) {
...
+	genericMaps, err := getGenericMaps(h.latestOutput)
+	if err != nil {
+		return nil, err
+	}
+	return genericMaps, nil
+}
+
+// DEPRECIATED - Use GetFilteredOutput or GetOutputCopy instead
+func (h *snapshotHistory) GetOutput() ([]byte, error) {
+	output, err := h.GetOutputCopy()
```
Sodman, line 178: Same Q as above, could we just wrap GetFilteredOutput with no filters and json format?

Sodman (commented):
Over LGTM, I'm a little concerned over potential performance implications for larger setups - I think we should test to make sure load times are still acceptable in that scenario, or have some way to bypass the filter-loading functionality by the user.

pkg/stats/index.go
josh-pritchard: It is yes. I am using it to get the list of clusters I can filter on and then I am using it to do the actual filtering. If this stops existing. Nothing will break. You just won't get cluster as an option when filtering

pkg/stats/io_snapshots.go
josh-pritchard: Yes

pkg/stats/index.go
```diff
@@ -8,8 +8,108 @@ import (
...
+	history SnapshotHistory
+}
+
+func NewIndex(history SnapshotHistory) Index {
+	return Index{
+		history: history,
+	}
+}
+
+// TODO - Use https://pkg.go.dev/embed once skv2 is upgraded to > go1.16
```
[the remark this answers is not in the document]
josh-pritchard: I can try that. Wasn't sure if we were okay with that but worth a try if its smooth

josh-pritchard:
@Sodman maybe when generating the html we wrap the capture of the filters in a try with a timeout. If it does not come back we render a simple html template (as it was previously). This way we serve up basic functionality still and protect against other, non performance, related issues as well.

pkg/stats/index.go
```diff
@@ -51,5 +58,112 @@ func Index(w http.ResponseWriter, r *http.Request) {
...
+	Error    error
+}
+
+func getTemplate(index Index, profiles []profile) (*template.Template, pageData) {
+	result := make(chan getIndexTemplateResult, 1)
+	go func() {
+		result <- getIndexTemplate(index, profiles)
+	}()
+	select {
+	case <-time.After(time.Duration(1.5 * float64(time.Second))):
```
josh-pritchard, line 81: Tested this with hard coded sleep

pkg/stats/index.go
```diff
@@ -51,5 +58,112 @@ func Index(w http.ResponseWriter, r *http.Request) {
...
+		log.Println("Timed out getting filters. Revert to fallback template.")
+		return getFallbackTemplate(index, profiles)
+	case result := <-result:
+		if result.Error != nil {
+			log.Printf("Unhandled error. Revert to fallback template. Error: %s", result.Error.Error())
```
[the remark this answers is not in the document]
josh-pritchard: whoops didn't even mean for that

Sodman (approved):
LGTM!

PR #354, opened 2022-07-14
fix(ReconcileSecret): add retry operations

gunnar-solo:


pkg/multicluster/watch/watcher.go
```diff
@@ -88,18 +85,30 @@ func (s *clusterWatcher) ListClusters() []string {
...
 		err := mgr.Start(ctx)
 		if err != nil {
-			contextutils.LoggerFrom(ctx).DPanicf("manager start failed for cluster %v: %v", clusterName, err)
+			contextutils.LoggerFrom(ctx).DPanicf("Manager creation/start failed for cluster %v: %v.  Trying again in %v seconds", clusterName, err, fib[retryAttempt])
+			time.Sleep(time.Duration(fib[retryAttempt]) * time.Second) // a bit of fibonacci backoff before next attempt
+
+			c.managers.delete(clusterName)        // remove failed manager from managers
+			c.handlers.RemoveCluster(clusterName) // remove failed manager from clusters
+
+			if retryAttempt >= len(fib) {
```
ilackarms, line 103: probably better to use the `retry` package here w/ exponential backoff

gunnar-solo: disclaimer:  _very plausible_ that I'm missing some basic understanding of `go`, which invalidates the below

-------

Here's an [example usage of the go retry package](https://pkg.go.dev/gopkg.in/retry.v1#example-Exponential):
```golang
// This example shows a retry loop that will retry an
// HTTP POST request with an exponential backoff
// for up to 30s.
strategy := retry.LimitTime(30*time.Second,
	retry.Exponential{
		Initial: 10 * time.Millisecond,
		Factor:  1.5,
	},
)
for a := retry.Start(strategy, nil); a.Next(); {
	if reply, err := http.Post("http://example.com/form", "", nil); err == nil {
		reply.Body.Close()
		break
	}
}
```

This doesn't seem to account for the what we're doing, here.  In particular:
* `err := mgr.Start(ctx)` needs to be in its own goroutine, since it is blocking _until it fails_
* `c.managers.set(clusterName, mgr, ctx, cancel)` needs to be called _after_ the goroutine is kicked off
* `c.managers.delete(clusterName)` needs to be called iff the blocking manager start fails

The difficult bit is that the `err` causes a retry can't _both_ be...
* waited for, to see if deploy succeeded
* not waited for, to internally register manager

gunnar-solo: if the ask is just "use exponential backoff" -- sure!  happy to make a quick swap.  If the ask is "use a specific package", I _think_ this is outside the scope of most simple stuff

nfuden: I think the suggestion is to use retry-go https://github.com/solo-io/skv2/blob/master/pkg/multicluster/register/registrant.go#L598

Also if the concern is that the retry itself isnt async you can always wrap it in a go func the way that the code is already doing

`
// This example shows a retry loop that will retry an
// HTTP POST request with an exponential backoff
// for up to 30s.
strategy := retry.LimitTime(30*time.Second,
	retry.Exponential{
		Initial: 10 * time.Millisecond,
		Factor:  1.5,
	},
)
go func(){ for a := retry.Start(strategy, nil); a.Next(); {
	if reply, err := http.Post("http://example.com/form", "", nil); err == nil {
		reply.Body.Close()
		break
	}
}}()
`

pkg/multicluster/watch/watcher.go
```diff
@@ -88,18 +86,33 @@ func (s *clusterWatcher) ListClusters() []string {
...
-	c.handlers.AddCluster(ctx, clusterName, mgr)
+		retry.Do(func() error {
+			ctx, cancel := context.WithCancel(contextutils.WithLoggerValues(context.WithValue(c.ctx, "cluster", clusterName), zap.String("cluster", clusterName)))
+			mgr, _ := manager.New(restCfg, c.optionsWithDefaults())
+
+			// add manager to managers+handlers.  It may fail to start and need to be removed ¯\_(ツ)_/¯
+			c.managers.set(clusterName, mgr, ctx, cancel)
+			c.handlers.AddCluster(ctx, clusterName, mgr)
+
+			err := mgr.Start(ctx) // blocking until error is thrown
```
elcasteel, line 105: Does this ever return `nil` (like a normal shutdown context)? Should we check before we write a scary error log?

PR #357, opened 2022-07-29
Verifier factory

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/356


pkg/verifier/verifier.go
```diff
@@ -96,8 +138,17 @@ func (c *cachedVerificationResponses) setCachedResponse(cluster string, gvk sche
...
+	c.lock.Lock()
+	defer c.lock.Unlock()
+	c.verifiedClusterResources = map[string]map[schema.GroupVersionKind]bool{}
+}
+
+// NewVerifier returns a new verifier
+// Deprecated: use VerifierFactory instead
 func NewVerifier(
 	ctx context.Context,
+	discoveryClient discovery.DiscoveryInterface,
```
yuval-k, line 151: this is a breaking change. is that intentional so you can fix all the invocations?

yuval-k (approved):

PR #369, opened 2022-08-17
Remove AdminService in template generation

fjork3:
Currently the AdminService abstraction is only being used within GME as a workaround to have a stats port correctly inserted into the generated Helm template; this PR removes that abstraction, to be replaced in GME generation with a `statsPort` field using existing overrides.


codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -190,6 +190,9 @@ spec:
   - name: [[ lower $port.Name ]]
     port: {{ $[[ $operatorVar ]].ports.[[ $port.Name ]] }}
   [[- end ]]
+  [[- if gt $[[ $operatorVar ]].statsPort 0 ]]
```
ashleywang1, line 193: This won't be available to the Operator since it's not described in the Operator spec. In this skv2 PR, it should be enough to just remove the AdminService. The statsPort changes will all happen in the GME PR.

ashleywang1 (changes requested):

codegen/templates/chart/operator-deployment.yamltmpl
fjork3: Gotcha, I thought that would need to be specified in the template being written out as well; is that just taken care of by the overrides?

ashleywang1: Yup! The ui-auth.yaml only exists in gloo-mesh-enterprise. The value it pulls in will be something like `glooMeshMgmtServer.statsPort`, which only needs to be defined in the gloo-mesh-enterprise values.yaml. We are essentially pulling this port completely out of the generated Operator deployment yamls.

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -204,39 +202,4 @@ spec:
 
 {{ include "skv2.utils.merge" (list . $[[ $operatorVar ]]ServiceOverrides "[[ $operator.Name ]].serviceSpec") }}
 {{- end }}
-
----
```
ashleywang1, line 208: This --- should not be removed.

ashleywang1 (approved):

PR #370, opened 2022-08-31
Fix generic client

EItanya:


elcasteel:
I think we also need fixes for the calls to `NewGenericClient` in the template code: https://github.com/solo-io/skv2/pull/364/files

PR #373, opened 2022-09-20
allow unknown fields

EItanya:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/372

### DIFF
--- a/codegen/templates/code/types/json.gen.gotmpl
+++ b/codegen/templates/code/types/json.gen.gotmpl
@@ -19,7 +19,10 @@ var _ = math.Inf
 
 var (
 	marshaller = &skv2jsonpb.Marshaler{}
-	unmarshaller = &jsonpb.Unmarshaler{}
+	unmarshaller = &jsonpb.Unmarshaler{
+		AllowUnknownFields: true,
+	}
+	strictUnmarshaller = &jsonpb.Unmarshaler{}
 )
 
 {{- range $resource := $.Resources }}
@@ -33,7 +36,11 @@ func (this *{{ $resource.Spec.Type.Name }}) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON is a custom unmarshaler for {{ $resource.Spec.Type.Name }}
 func (this *{{ $resource.Spec.Type.Name }}) UnmarshalJSON(b []byte) error {
+{{- if $resource.StrictUnmarshal }}
+	return strictUnmarshaller.Unmarshal(bytes.NewReader(b), this)
+{{- else }}
 	return unmarshaller.Unmarshal(bytes.NewReader(b), this)
+{{- end }}
 }
 {{- end }}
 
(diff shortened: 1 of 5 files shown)


codegen/templates/code/types/json.gen.gotmpl
```diff
@@ -33,7 +36,11 @@ func (this *{{ $resource.Spec.Type.Name }}) MarshalJSON() ([]byte, error) {
 
 // UnmarshalJSON is a custom unmarshaler for {{ $resource.Spec.Type.Name }}
 func (this *{{ $resource.Spec.Type.Name }}) UnmarshalJSON(b []byte) error {
+{{- if $resource.StrictUnmarshal }}
```
nfuden, line 39: I like this but do we also feel like adding something in utils or on the codegenned method set to identify the extra fields?

like loose unmarshal, remarshal to json and then compare or some such.

codegen/templates/code/types/json.gen.gotmpl
```diff
@@ -19,7 +19,10 @@ var _ = math.Inf
 
 var (
 	marshaller = &skv2jsonpb.Marshaler{}
-	unmarshaller = &jsonpb.Unmarshaler{}
+	unmarshaller = &jsonpb.Unmarshaler{
+		AllowUnknownFields: true,
```
jenshu, line 23: if we made a breaking api change (we shouldn't do that, but let's say that we did), and say changed the type of a field from a string to a boolean, and then we tried to unmarshal a resource that was still using a string value, would unmarshaling fail, or would that field be treated as an 'unknown' field and unmarshaling would succeed? (i would like to find a way for unmarshaling to succeed even on breaking api changes like this)

EItanya: TBH I'm not sure, that's not the purpose of this change. We need to remove an unecessary status field in GM, and this will allow that.

codegen/templates/code/types/json.gen.gotmpl
EItanya: I suppose, as mentioned in the comment below, this change has a very specific purpose, so we do not need that. If it becomes necessary I'm happy to help add it.

codegen/templates/code/types/json.gen.gotmpl
ilackarms: unmarshaling would fail in that case.

allow unknown field specificallly handles the case when an unknown field is discovered.

your example is a known field which will fail to unmarhsal with error akin to `cannot unmarshal field of type X into type Y`

codegen/templates/code/types/json.gen.gotmpl
ilackarms: tbh i don't think we have any need for this at all. however if we feel like leaving the option for some unforseen edge case, i think the simplicity here is good enough

PR #378, opened 2022-09-22
Remove panic() calls

Bslabe123:
BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/7174


codegen/kuberesource/crd.go
```diff
@@ -175,7 +177,8 @@ func CustomResourceDefinition(
 		// hashstructure of the crd spec:
 		specHash, err := hashstructure.Hash(crd.Spec, nil)
 		if err != nil {
-			panic(err)
+			contextutils.LoggerFrom(nil).DPanic(err)
```
ben-taussig-solo, line 180: Given that this is in codegen, I think we might prefer to panic here. I think codegen failing loudly is preferable to potentially returning an incorrect or undesirable result

contrib/README.md
```diff
@@ -19,7 +19,8 @@ const (
 var Sets = func() model.CustomTemplates {
 	templateContents, err := templatesBox.FindString(SetCustomTemplatePath)
 	if err != nil {
-		panic(err)
```
ben-taussig-solo, line 22: I think we want to revert this -- I think the intended behavior is that we panic if we can't find the template in the specified location

contrib/codegen/funcs/funcs.go
```diff
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/solo-io/go-utils/contextutils"
```
ben-taussig-solo, line 9: I think we can revert the changes here -- I don't think it's an issue if codegen panics

contrib/codegen/templates/sets/sets.gotmpl
```diff
@@ -136,7 +137,7 @@ func (s *{{ $kindLowerCamel }}Set) Insert(
         {{ $kindLowerCamel }}List ...*{{ $import_prefix }}.{{ $resource.Kind }},
 ) {
     if s == nil {
-        panic("cannot insert into nil set")
+        contextutils.LoggerFrom(nil).DPanic("cannot insert into nil set")
```
ben-taussig-solo, line 140: Same as above -- I think we can revert the changes in this file because it isn't problematic for codegen to panic

ben-taussig-solo (commented):
Left a couple of comments -- Essentially, I question whether we should be making changes to codegen, as I think it might be preferable for codegen to panic in the case of bad input.

contrib/custom_top_level_templates.go
```diff
@@ -96,10 +98,12 @@ const (
 // Returns the template for generating input snapshots.
 func InputSnapshot(params SnapshotTemplateParameters) model.CustomTemplates {
 	templateContentsBytes, err := ioutil.ReadFile(templatesDir + InputSnapshotCustomTemplatePath)
+	var templateContents string
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(nil).DPanic(err)
+		templateContents = ""
```
sam-heilbron, line 104: This sets the templatecontents, and then the next line (outside the conditional) overwrite it. Do we need to do this or should we be returning an error here?

contrib/codegen/templates/sets/sets.gotmpl
```diff
@@ -6,6 +6,7 @@ import (
     {{ $import_prefix }} "{{ group_import_path $ }}"
 
     "github.com/rotisserie/eris"
+    "github.com/solo-io/go-utils/contextutils"
```
sam-heilbron, line 9: Why is this import needed?

changelog/v0.23.5/remove-panic.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/gloo/issues/7174
+    resolvesIssue: true
```
sam-heilbron, line 4: false perhaps? Since we'll need to pull in these changes to gloo, I assume this won't be true until we have made the change in all repos

pkg/events/watcher.go
```diff
@@ -106,7 +108,7 @@ func (w *eventWatcher) Reconcile(ctx context.Context, request reconcile.Request)
 			return reconcile.Result{}, err
 		}
 	default:
-		panic("invalid event")
+		return reconcile.Result{}, eris.Errorf("invalid event")
```
sam-heilbron, line 111: Do you think it makes sense to include any context around the event type so that if it occurs we can more easily debug it?

sam-heilbron:
@Sodman can you or someone from the Mesh team check out this to make sure we're on the same page about the implications?

ben-taussig-solo (approved):
Looks good to me. For context, I've pushed to keep panics in codegen, under the assumptions that panicking on bad input is desired behavior in codegen/tests/other code which doesn't operate on the application path.

Sodman:
If I understand correctly, these changes only affect codegen, and not any runtime, right?

My take here is that I'd like it if codegen explodes loudly and abruptly if there's something wrong. If we are going this route though, I think we should still return early with an error though, as otherwise we'll probably just log a single line in what's already a very output-verbose cli command - I'm worried about codegen accidentally being "broken" but the CLI command suddenly returning a happy exit 0 status and the developer not noticing  / committing the code to a PR.

If we're trying to remove `panic()`'s to keep a linter happy or some such though, I'm sure we can change our codegen on the mesh side make targets to enable devlogger, which should still result in panics on our side when we run this if we want them.

contrib/custom_top_level_templates.go
```diff
@@ -113,10 +117,14 @@ const (
 // Returns the template for generating input snapshots.
 func InputSnapshotManualBuilder(params SnapshotTemplateParameters) model.CustomTemplates {
 	templateContentsBytes, err := ioutil.ReadFile(templatesDir + InputSnapshotManualBuilderCustomTemplatePath)
+	var templateContents string
 	if err != nil {
-		panic(err)
+		if err != nil {
```
sam-heilbron, line 122: I think this duplicate error check is unnecessary

contrib/custom_top_level_templates.go
```diff
@@ -136,7 +144,7 @@ func InputReconciler(params SnapshotTemplateParameters) model.CustomTemplates {
 	}
 	templateContentsBytes, err := ioutil.ReadFile(templatesDir + templatePath)
 	if err != nil {
-		panic(err)
+		contextutils.LoggerFrom(nil).DPanic(err)
```
sam-heilbron, line 147: Should we return error as well?

contrib/custom_top_level_templates.go
EItanya: Same comment as above

contrib/custom_top_level_templates.go
EItanya: Same comment

contrib/codegen/templates/sets/sets.gotmpl
Bslabe123: This was leftover from a previous change, fixed

Bslabe123:
See https://solo-io-corp.slack.com/archives/CHAC8587L/p1663947139651859?thread_ts=1663946678.931739&cid=CHAC8587L

PR #379, opened 2022-09-22
don't pass k8s through camel case

josh-pritchard:
This could impact more fields so eventually should come up with more full proof solution


codegen/proto/options.go
```diff
@@ -256,3 +256,13 @@ func getFieldOptionOpenAPIValidationDisabled(field *descriptor.FieldDescriptorPr
 	}
 	return false, nil
 }
+
+func lowerCamelName(field FieldOptions) string {
+	name := field.Field.GetName()
+	// Hack because strcase.ToLowerCamel is not converting string the same as the `cue` lib is
+	if name == "k8s" {
```
Sodman, line 263: How confident are we that this string is always already lowercase? Should we do a name.ToLower() here?

josh-pritchard: This is right I confirmed this change fixes the generated CRDs locally

codegen/proto/options.go
```diff
@@ -236,23 +238,34 @@ func ParseOptions(fileDescriptors []*collector.DescriptorWithPath) (Options, err
...
+func getFieldOptionValues(field *descriptor.FieldDescriptorProto) (bool, string, error) {
 	cueOptRaw, err := proto.GetExtension(field.Options, cue.E_Opt)
 	if err == nil {
 		cueOpt, ok := cueOptRaw.(*cue.FieldOptions)
 		if !ok {
-			return false, eris.Errorf("internal error: invalid option type %T expecting *cueproto.FieldOptions", cueOpt)
+			return false, "", eris.Errorf("internal error: invalid option type %T expecting *cueproto.FieldOptions", cueOpt)
 		}
-		return cueOpt.DisableOpenapiValidation, nil
+		return cueOpt.DisableOpenapiValidation, cueOpt.JsonFieldNameOverride, nil
```
ilackarms, line 259: nit: instead of returning each option as a separate arg, what about a struct with all opts?

ilackarms (approved):

EItanya (approved):

PR #393, opened 2022-11-14
Migrate away from metadata.ClusterName use for k8s 1.25 support

conradhanson:
- Go dependencies updated to k8s 1.25
- Added explicit support for k8s 1.24 where interface for accessing clusterName field changes
- single method GetClusterName() handles all possible interfaces for actually getting value

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/5158


api/core/v1/core.proto
```diff
@@ -39,6 +39,9 @@ message ClusterObjectRef {
 
     // name of the cluster in which the resource exists
     string cluster_name = 3;
+
+    // annotations of the resource being referenced
+    map<string, string> annotations = 4;
```
EItanya, line 44: This feels a little strange to me. From the reference perspective I think we should still just have a `cluster_name`. No matter how it's represented in kubernetes.

conradhanson: Agreed. I am removing it upon reconsideration.

At first, I did this because I found case in GME where ClusterObjectRef was being passed into a function that used a ClusterResourceId input param. So to create parity we could add that field to the proto so a GetAnnotations method would be created. But to date, this is the only discrepancy I've found, so its an easy fix to that function and its use, instead of polluting the protos.

conradhanson: Actually this is still very much an issue throughout GME.

yuval-k:
i also have  a breaking change https://github.com/solo-io/skv2/pull/394
are you planning to merge this soon?

conradhanson:
> i also have a breaking change #394 are you planning to merge this soon?

I am not. Still have work to do on this.

contrib/pkg/sets/sets.go
```diff
@@ -34,18 +34,14 @@ func Key(id ezkube.ResourceId) string {
...
-		return b.String()
-	}
 	b.WriteString(id.GetName())
 	b.WriteString(separator)
 	b.WriteString(id.GetNamespace())
 	b.WriteString(separator)
+	if clusterId, ok := id.(ezkube.ClusterResourceId); ok {
+		b.WriteString(separator)
+		b.WriteString(ezkube.GetClusterName(clusterId))
+	}
```
EItanya, line 44: What if something is passed in here that fulfills the old interface? e.g. `GetClusterName()`

conradhanson: As it stands, only the clusterName annotation would be checked, which would result in the clusterName effectively not propagating. So I should probably add another helper function in ezkube package to handle that case too.

codegen/kuberesource/crd_test.go
```diff
@@ -40,9 +40,9 @@ var _ = Describe("Crd", func() {
 			o, err := CustomResourceDefinitions(grp)
 			Expect(err).NotTo(HaveOccurred())
 			Expect(o).To(HaveLen(1))
-			// note: we intentionally provide the "5770277ee8c9bc92" hash in the test, as it shouldn't change
+			// note: we intentionally provide the "b6ec737002f7d02e" hash in the test, as it shouldn't change
```
jenshu, line 43: do you know why this hash (and all the hashes in the CRDs) changed?

pkg/multicluster/kubeconfig/kubeconfig_secret_test.go
```diff
@@ -32,7 +32,6 @@ preferences: {}
 users:
 - name: test-name
   user:
-    as-user-extra: null
```
jenshu, line 35: why were a bunch of files under `pkg/multicluster` deleted? are we sure we don't need those anymore?
and why was this line removed?

conradhanson: That is a great question that I don't immediately have the answer to. Looks like those files were deleted as part of Eitan's initial commits. @EItanya do you have context on why those files were removed?

codegen/kuberesource/crd_test.go
conradhanson: Not entirely sure, I can try and track down why. 
My initial thought is potentially because of related go dependency upgrades on packages that are used to generate the CRDs. Particularly `code-generator`, but that is a complete guess rn.

contrib/pkg/sets/sets.go
```diff
@@ -34,18 +34,21 @@ func Key(id ezkube.ResourceId) string {
...
 	b.WriteString(separator)
+	if clusterId, ok := id.(ezkube.ClusterResourceId); ok {
+		b.WriteString(separator)
+		// handle the possibility that clusterName could be set either as an annotation (new way)
+		// or as a field (old way pre-k8s 1.25)
+		clusterNameByAnnotation := ezkube.GetClusterName(clusterId)
+		if clusterNameByAnnotation != "" {
+			b.WriteString(clusterNameByAnnotation)
+		} else {
+			b.WriteString(ezkube.GetDeprecatedClusterName(id))
```
EItanya, line 49: I don't think this will work as the other object does not conform to `ezkube.ClusterResourceId`

contrib/pkg/sets/sets.go
```diff
@@ -34,18 +34,25 @@ func Key(id ezkube.ResourceId) string {
...
 	b.WriteString(id.GetNamespace())
 	b.WriteString(separator)
+	// handle the possibility that clusterName could be set either as an annotation (new way)
+	// or as a field (old way pre-k8s 1.25)
+	if clusterId, ok := id.(ezkube.ClusterResourceId); ok {
+		clusterNameByAnnotation := ezkube.GetClusterName(clusterId)
+		if clusterNameByAnnotation != "" {
+			b.WriteString(separator)
+			b.WriteString(clusterNameByAnnotation)
+		} else if deprecatedClusterId, ok := id.(interface{ GetClusterName() string }); ok {
```
nfuden, line 48: why are we definining the interface in line? also why do we have a double else if here?

conradhanson: 1. so we don't have to expose the deprecatedClusterId interface defined in ezkube, we can't really define it in this file since we need it for ConvertRefToId(...)
2. double else if bc there is a possibility that a pre-k8s-1.25 object either has the annotation or the clusterName field defined, so we should handle both cases (inside first if). If the object is of ClusterObjectRef type, then it will not match the new ClusterResourceId interface (first if), and thus needs to be handled as follow-on else if statement.

contrib/pkg/sets/sets.go
```diff
@@ -34,18 +34,25 @@ func Key(id ezkube.ResourceId) string {
...
+			b.WriteString(separator)
+			b.WriteString(deprecatedClusterId.GetClusterName())
+		}
+	} else if deprecatedClusterId, ok := id.(interface{ GetClusterName() string }); ok {
+		b.WriteString(separator)
+		b.WriteString(deprecatedClusterId.GetClusterName())
+	}
```
nfuden, line 55:
```suggestion
		if clusterId, ok := id.(ezkube.ClusterResourceId); ok {
		clusterNameByAnnotation := ezkube.GetClusterName(clusterId)
		if clusterNameByAnnotation != "" {
			b.WriteString(separator)
			b.WriteString(clusterNameByAnnotation)
			return b.String()
		}

	}
	// handle the possibility that clusterName could be set either as an annotation (new way)
	// or as a field (old way pre-k8s 1.25)
	if deprecatedClusterId, ok := id.(interface{ GetClusterName() string }); ok {
		b.WriteString(separator)
		b.WriteString(deprecatedClusterId.GetClusterName())
	}
```

nfuden:
Observation: I love that this is on release v0.25 and supports k8s 1.25 :)

contrib/pkg/sets/sets.go
```diff
@@ -34,18 +34,24 @@ func Key(id ezkube.ResourceId) string {
...
 	b.WriteString(id.GetName())
 	b.WriteString(separator)
 	b.WriteString(id.GetNamespace())
 	b.WriteString(separator)
+	// handle the possibility that clusterName could be set either as an annotation (new way)
+	// or as a field (old way pre-k8s 1.25)
+	if clusterId, ok := id.(ezkube.ClusterResourceId); ok {
+		clusterNameByAnnotation := ezkube.GetClusterName(clusterId)
+		if clusterNameByAnnotation != "" {
+			b.WriteString(separator)
```
conradhanson, line 46: This separator is extraneous, correct? Same with the one defined on line 47, since we already have one defined on line 40. 
This cropped up in a graphql unit test where there was a double dot between ns and cluster parts of the key.

EItanya:
These changes LGTM, would love a last look from @nfuden and @jenshu as well.

EItanya (approved):

pkg/ezkube/resource_id.go
```diff
@@ -88,18 +95,26 @@ func ConvertRefToId(ref deprecatedClusterResourceId) ClusterResourceId {
...
-	if id.GetAnnotations() == nil {
-		return ""
+	annotations := id.GetAnnotations()
+	deprecatedClusterName := getDeprecatedClusterName(id)
+
+	if annotations == nil {
+		return deprecatedClusterName
+	} else if annotations[ClusterAnnotation] == "" {
+		return deprecatedClusterName
 	}
```
jenshu, line 115:
```suggestion
	if annotations == nil || annotations[ClusterAnnotation] == "" {
		return getDeprecatedClusterName(id)
	}
```
nit: only call the deprecated func if needed

changelog/v0.26.2/support-k8s-1.25.yaml
```diff
@@ -0,0 +1,6 @@
+changelog:
+    - type: NON_USER_FACING
```
jenshu, line 2: is this no longer a breaking change?

conradhanson: not since we merged in the breaking changes that became v.0.26, I think?

pkg/ezkube/resource_id.go
conradhanson: i like it!

EItanya (approved):

changelog/v0.26.2/support-k8s-1.25.yaml
nfuden: nah its breaking

jenshu: i think we are going to have issues with gloo edge LTS branches picking up future skv2 patch fixes if this isn't marked as a breaking change. mostly concerned about the go mod updates to 0.25 k8s api

jenshu (approved):

EItanya (approved):

PR #402, opened 2022-12-12
Feat/k8s no clustername

nfuden:
This pr is split from https://github.com/solo-io/skv2/pull/393

This PR should not update go mod and purely enable preexisting clustername usages to instead work via the annotation storage method.
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/5158


pkg/resource/mocks/client.go
```diff
@@ -38,17 +38,22 @@ func (m *MockClient) EXPECT() *MockClientMockRecorder {
 }
 
 // Get mocks base method
-func (m *MockClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object) error {
+func (m *MockClient) Get(ctx context.Context, key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
```
jenshu, line 41: also i think this can be reverted

crds/multicluster.solo.io_v1alpha1_crds.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    crd.solo.io/specHash: ad9e1dd2ecf8891d
+    crd.solo.io/specHash: 396fd84315636ecb
```
jenshu, line 7: was this change auto-generated?

pkg/multicluster/register/registrant.go
```diff
@@ -594,26 +594,64 @@ func (c *clusterRegistrant) getTokenForSa(
...
+
+	// create a secret name format. This is similar to what was previously done
+	// but after the k8s upgrade we no longer have knowledge of what was previously
+	// autogenerated by k8s.
+	// TODO(nfuden): move away from these long lived connections as they are
+	// likely to be deprecated even more in the future.
+	secretName := fmt.Sprintf("%s-secret", sa.Name)
+
+	// see if there is a secret
+	_, err = remoteSecretClient.GetSecret(ctx, client.ObjectKey{Name: secretName, Namespace: saRef.Namespace})
```
jenshu, line 618: this should be moved into the retry loop, and why is there a separate call to GetSecret in the retry below?

nfuden: At this point we do not believe there exists a secret already. If there is we should not destroy it by our upsert.
This is why we dont have this in the loop.

The loop get secret is to make sure the upsert and k8s controller update has taken place

jenshu (approved):

conradhanson (approved):

PR #403, opened 2023-01-04
Fix ratelimit key

jenshu:
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/404

### DIFF
--- a/contrib/pkg/input/input_reconciler.go
+++ b/contrib/pkg/input/input_reconciler.go
@@ -43,6 +43,8 @@ type inputReconciler struct {
 	singleClusterReconcileFunc SingleClusterReconcileFunc
 }
 
+const keySeparator = "/"
+
 // Note(ilackarms): in the current implementation, the constructor
 // also starts the reconciler's event processor in a goroutine.
 // Make sure to cancel the parent context in order to ensure the goroutine started here is gc'ed.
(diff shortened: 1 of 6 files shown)


contrib/pkg/input/input_reconciler.go
```diff
@@ -110,31 +111,42 @@ func (r *inputReconciler) reconcileEventsForever(reconcileInterval time.Duration
...
 	}
-	defer r.queue.Done(key)
+	defer r.queue.Done(queueItem)
 
-	var isRemoteCluster bool
+	// get the string representation of the queue item
+	key, ok := queueItem.(string)
+	if !ok {
+		contextutils.LoggerFrom(r.ctx).Errorw("got a work queue item of non-string type", "item", queueItem)
+		r.queue.Forget(queueItem)
```
nfuden, line 124: should we return after the forgets?

conradhanson: new line 159 is showing that this is the default behavior so I'd think we would want to

jenshu: ha yeah good catch, both places where we error, we should return

conradhanson (changes requested):
LGTM just needs a changelog

nfuden (approved):

conradhanson (approved):
expects on exact error strings in tests are usually brittle (should prob compare by error type instead), but since this code shouldn't be changing much I think it's fine.

nfuden (approved):

PR #406, opened 2023-01-31
Helm Gen Enhancements

josh-pritchard:
Doesn't look like much now that I have it done but was a pain to get right. Result is a clean merge of the value file vs. grouping them per operator. This allows for nesting.

- Allow nested operators
- Don't add empty template files when no operators are present

Usage can be seen here https://github.com/solo-io/gloo-mesh-enterprise/pull/7364

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/407


codegen/render/funcs.go
```diff
@@ -75,6 +76,40 @@ func makeTemplateFuncs(customFuncs template.FuncMap) template.FuncMap {
...
+			}
+			if o.ValuePath != "" {
+				splitPath := strings.Split(o.ValuePath, ".")
+				for _, p := range splitPath {
+					opValues = map[string]interface{}{
+						strcase.ToLowerCamel(p): opValues,
+					}
+				}
+			}
+			return fromYAML(toYAML(opValues))
```
npolshakova, line 92: Is this double conversion just to check if the values are valid yaml?

codegen/model/chart.go
```diff
@@ -150,7 +153,8 @@ func (c Chart) BuildChartValues() values.UserHelmValues {
 		}
 
 		helmValues.Operators = append(helmValues.Operators, values.UserOperatorValues{
-			Name: operator.Name,
+			Name:      operator.Name,
+			ValuePath: operator.ValuePath,
```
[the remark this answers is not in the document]
josh-pritchard: They don't overwrite each other, value path represents the path at which the operator values will be written. That way we can nest operators in the values API. Maybe its not the best naming, any ideas?

codegen/render/funcs.go
josh-pritchard: quick and dirty method to get it to `map[string]interface{}` instead of the structs so the built in mergeOverwrite function will work in the template

codegen/model/chart.go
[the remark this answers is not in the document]
npolshakova: Ah okay that makes sense! I think the example in the comment makes in clear ("agent" would place an operator at "agent.<operatorName>")

npolshakova (dismissed):

codegen/render/funcs.go
```diff
@@ -193,6 +237,89 @@ func fromYAML(str string) map[string]interface{} {
 	return m
 }
 
+func toNode(v interface{}) goyaml.Node {
```
josh-pritchard, line 240: @albertlockett you will be able to integrate your comments addition code here.

albertlockett: @josh-pritchard this looks great!

I think when we integrate the comment stuff, since we'll want to make it opt-in, I'll end up passing an additional argument to this to control commenting (e.g. is commenting the yaml enabled, what is our line break length is).

So w/ that said, I wonder if we'd want to eventually have `get_operator_custom_values` and `get_operator_values` return map[string]interface{} and then we'll pipe that into get node ?

```diff
- [[- $customValues = mergeNodes $customValues (get_operator_values $operator) (get_operator_custom_values $operator) ]]
+ [[- $customValues = mergeNodes $customValues (toNode (get_operator_values $operator) $config ) (toNode (get_operator_custom_values $operator) $config)  ]]
```

josh-pritchard: yeah that makes sense to me, seriously feel free to adjust anything that makes sense here when you bring in your changes.

albertlockett (approved):

npolshakova (approved):

PR #408, opened 2023-02-06
prevent data race on clustername annotation

conradhanson:
upgrading to go 1.20 on gme has exposed a data race when GetClusterName and SetClusterName are called simultaneously and try to access/set annotations map.

Added a mutex that is shared by both functions to prevent the race from being caused by these functions, note this doesn't fully prevent races happening on the annotations map.

### DIFF
--- a/changelog/v0.27.1/prevent-clustername-data-race.yaml
+++ b/changelog/v0.27.1/prevent-clustername-data-race.yaml
@@ -1,4 +1,6 @@
 changelog:
     - type: FIX
+      issueLink: https://github.com/solo-io/gloo-mesh-enterprise/issues/7412
+      resolvesIssue: false
       description: |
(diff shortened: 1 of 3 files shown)


conradhanson:
would like to merge this after this [gme pr](https://github.com/solo-io/gloo-mesh-enterprise/pull/7427) has a fully passing ci run

conradhanson:
[Get:SetClusterName data race.log](https://github.com/solo-io/skv2/files/10668270/Get.SetClusterName.data.race.log)

jenshu (approved):

EItanya (commented):
I understand the intention of this fix, but I don't love the idea of a global lock to be honest. I looked at the logs and it's not even clear to me that the code causing that race needs to run at all. Let's catch up this AM

PR #413, opened 2023-02-15
add ability to generate chart schema.json

albertlockett:
Adds the ability to generate the json schema for a chart. Users will do this by adding the `JsonSchema` field to the chart definition as shown here:

https://github.com/solo-io/skv2/blob/33b1815a064e71bd033f148d63c323d5e60d95e4/codegen/cmd_test.go#L309

BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/7186

### DIFF
--- a/codegen/render/funcs.go
+++ b/codegen/render/funcs.go
@@ -575,6 +575,7 @@ func fromJSON(str string) map[string]interface{} {
 func toJSONSchema(values values.UserHelmValues) string {
 	reflector := createJsonSchemaReflector(values)
 	schema := new(jsonschema.Schema)
+	schema.Version = "https://json-schema.org/draft/2020-12/schema"
 
 	// add json schema properties from the chart's custom values
 	if values.CustomValues != nil {
@@ -857,6 +858,7 @@ func jsonSchemaForOperator(
 	o *values.UserOperatorValues,
 ) *jsonschema.Schema {
 	schema := reflector.Reflect(o.Values)
+	schema.Version = "" // will be set on parent object
 
 	// if the opeartor defines custom values, add the properties to the schema
 	if o.CustomValues != nil {
--- a/codegen/render/funcs.go
+++ b/codegen/render/funcs.go
@@ -9,13 +9,19 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/invopop/jsonschema"
 	"github.com/solo-io/skv2/codegen/model/values"
 	"github.com/solo-io/skv2/codegen/util/stringutils"
+	"google.golang.org/protobuf/types/known/structpb"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/strings/slices"
 
 	"github.com/BurntSushi/toml"
 	"github.com/Masterminds/sprig/v3"
 	"github.com/golang/protobuf/protoc-gen-go/descriptor"
+	"github.com/iancoleman/orderedmap"
 	"github.com/iancoleman/strcase"
 	"github.com/solo-io/skv2/codegen/model"
 	"github.com/solo-io/skv2/codegen/util"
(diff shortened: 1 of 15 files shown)


codegen/render/render.go
```diff
@@ -12,8 +12,13 @@ type Resource = model.Resource
...
+}
+
 type OutFile struct {
-	Path       string
-	Permission os.FileMode
-	Content    string // set by Renderer
+	Path           string
+	Permission     os.FileMode
+	Content        string         // set by Renderer
+	HeaderOverride HeaderOverride // overrides the defauld the default header of the outfile writer
```
josh-pritchard, line 23: typo

codegen/render/funcs.go
```diff
@@ -562,6 +571,341 @@ func fromJSON(str string) map[string]interface{} {
 	return m
 }
 
+// toJSONSchema generates the json schema for the values of the helm chart
+func toJSONSchema(values values.UserHelmValues) string {
+	reflector := createJsonSchemaReflector(values)
+	schema := new(jsonschema.Schema)
+	schema.Version = "https://json-schema.org/draft/2020-12/schema"
```
josh-pritchard, line 578: Where did this come from? Maybe a comment or link?

codegen/render/funcs.go
```diff
@@ -562,6 +571,341 @@ func fromJSON(str string) map[string]interface{} {
...
+		}
+
+		return nil
+	}
+}
+
+// initialize a new instance of the jsonschema Reflector with configured to
+// create json schema for chart.
+func createJsonSchemaReflector(values values.UserHelmValues) *jsonschema.Reflector {
+	// we might want to pass configuration ot this from values.UserHelmValues
```
josh-pritchard, line 700:
```suggestion
	// we might want to pass configuration of this from values.UserHelmValues
```

codegen/render/funcs.go
```diff
@@ -562,6 +571,341 @@ func fromJSON(str string) map[string]interface{} {
...
+	var buf bytes.Buffer
+	err := json.Indent(&buf, []byte(src), "", "  ")
+	if err != nil {
+		panic(err)
+	}
+	return buf.String()
+}
+
+// reverse mutates a slice of strings to reverse the order.
+func reverse(s []string) {
```
josh-pritchard, line 903: one would think golang standard lib would include this....not that I could find it.

codegen/writer/writer.go
```diff
@@ -49,7 +49,9 @@ func (w *DefaultFileWriter) WriteFiles(files []render.OutFile) error {
 			commentPrefix = "#"
 		}
 
-		if w.Header != "" {
+		if file.HeaderOverride != nil {
```
josh-pritchard, line 52: nit: couldn't we disable the header based on file type? I am fine with it as is too.

josh-pritchard (commented):
only a few nits: This looks fantastic and being an opt-in feature lets get this in so we can try it out

codegen/render/funcs.go
albertlockett: yeah I googled for quite a while and I couldn't find something that does it :(

codegen/writer/writer.go
albertlockett: yeah for sure good call, I agree w/ that in hindsight

At the time I thought this HeaderOverride solution might be more flexible but given that this one json file is the only use case and the added complexity I'm inclined to just disable header for json as well. it aligns better w/ the current pattern of choosing header by file type.

albertlockett: fixed, I removed this

codegen/render/funcs.go
albertlockett: fixed, I got this from the jsonschema package.

codegen/render/render.go
albertlockett: fixed I removed this.

codegen/render/funcs.go
fernandosanchezs: `Implementation is left to the reader.`

puertomontt: wow

josh-pritchard (dismissed):
LGTM! Great work!

puertomontt (approved):

josh-pritchard (approved):

Issue #419, opened 2023-02-21
Bump controller-runtime to 0.14.x


birkland:
Istio 1.17+ bumped their controller-runtime dependency from 0.13.x to 0.14.x.  Unfortunately, the controller-runtime API contains at least one breaking change that would impact skv2.  In order for any repo to bump their Istio dependencies to 1.17 or newer, the `controller-runtime` also needs to be updated.   Bump the skv2 `controller-runtime` dependency and deal with the API breakages in order to allow eventual compatibility with Istio v1.17.x dependencies

PR #420, opened 2023-02-21
Bump controller-runtime to v0.14.x

birkland:
`controller-runtime` introduced breaking changes in the status API from v0.13.x -> v0.14.x.  Unfortunately, upstream Istio is now aligned with v1.14.x, so bumping skv2 is a necessary prerequisite for repos that want to include `istio.io/istio@v1.17.0` or newer.

This PR bumps the dependencies, updates API signatures, and updates codegen templates to comply.  I don't know why codegen updated schema json files.
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/419


changelog/v0.30.0/update-controller-runtime.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: BREAKING_CHANGE
```
jenshu, line 2: could you also list the DEPENDENCY_BUMPS, particularly k8s.io/api and controller-runtime?

codegen/templates/code/types/clients.gotmpl
```diff
@@ -129,10 +129,10 @@ type {{ $resource.Kind }}Writer interface {
 type {{ $resource.Kind }}StatusWriter interface {
     // Update updates the fields corresponding to the status subresource for the
     // given {{ $resource.Kind }} object.
-    Update{{ $resource.Kind }}Status(ctx context.Context, obj *{{ $import_prefix }}{{ $resource.Kind }}, opts ...client.UpdateOption) error
+    Update{{ $resource.Kind }}Status(ctx context.Context, obj *{{ $import_prefix }}{{ $resource.Kind }}, opts ...client.SubResourceUpdateOption) error
```
jenshu, line 132: do you know why these had to be renamed? as far as i can tell, `UpdateOption` and `PatchOption` both still exist in controller-runtime. do they behave differently in the newest controller-runtime (e.g., are they enforcing that we call the 'subresource' methods on statuses now)?

birkland: From https://github.com/kubernetes-sigs/controller-runtime/pull/2072
> Use distinct options for subresource Update and Patch which makes this a breaking change

So I can't comment on _why_ these API breakages had to happen other than link to the upstream pull request, but I can say that while `UpdateOption` and PatchOption` still exist, but now can only be applied to resources that aren't subresources.

changelog/v0.30.0/update-controller-runtime.yaml
birkland: Yeah, those seem reasonable.  All were a consequence of pulling in controller-runtime.

codegen/test/chart-no-desc/values.schema.json
```diff
@@ -8076,6 +8194,67 @@
                                   "type": "null"
                                 }
                               ]
+                            },
+                            "schedulingGates": {
```
jenshu, line 8198: looks like these and other schema.json updates came from new fields being added to the pod spec in the latest k8s api

jenshu (approved):

ashleywang1:
The GME PR that uses this branch is [here](https://github.com/solo-io/gloo-mesh-enterprise/pull/8222) and has trouble compiling.

t-edris:
what lol
Changelog bot failed before the last commit because it was v0.30.0, so I moved it to v0.29.6, but now it's failing because it's _not_ v0.30.0?

go.mod
```diff
@@ -39,63 +39,53 @@ require (
...
-	k8s.io/utils v0.0.0-20220728103510-ee6ede2d64ed
-	sigs.k8s.io/controller-runtime v0.13.1
+	k8s.io/api v0.26.4
+	k8s.io/apiextensions-apiserver v0.26.4
+	k8s.io/apimachinery v0.26.4
+	k8s.io/client-go v0.26.4
+	k8s.io/code-generator v0.26.4
+	k8s.io/klog/v2 v2.80.1
+	k8s.io/utils v0.0.0-20221128185143-99ec85e7a448
+	sigs.k8s.io/controller-runtime v0.14.4
```
nfuden, line 55: https://github.com/kubernetes-sigs/controller-runtime/blob/main/go.mod#L3
We should bump our go version to align with this dependency

t-edris: Okay, will do. Is that something that would be called out in the Changelog as well?

ashleywang1 (approved):

marcogschmidt (approved):

PR #423, opened 2023-03-01
values.yaml inline docs edge cases

albertlockett:
handles two new cases when generating inline docs values which were not supported before:
- when an struct has another struct inlined and the inlined struct has fields that are the same name as the type
- when a child struct is completely initialized to 0 values
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/424


josh-pritchard (approved):
Works perfect, thanks for jumping on this!

puertomontt (approved):

PR #431, opened 2023-03-09
Ginkgo v2

bewebi:
We're working on [bumping Portal to k8s 1.25](https://github.com/solo-io/dev-portal/issues/2613)
This requires bumping to certain internal deps which support k8s 1.25, however some of those require ginkgo v2
Therefore we are also working on bumping Portal to ginkgo v2
Portal relies on skv2 packages which rely on ginkgo v1
Therefore we must release a version of skv2 which uses ginkgo v2


go.mod
```diff
@@ -23,8 +23,8 @@ require (
 	github.com/lithammer/dedent v1.1.0
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/hashstructure v1.0.0
-	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.20.1
+	github.com/onsi/ginkgo/v2 v2.5.0
```
jenshu, line 26: we use v2.8.1 in gloo/solo-projects, should we upgrade to that?

bewebi: I don't see why not...the dep that led to requiring ginko v2 was solo-kit which [uses v2.5.0](https://github.com/solo-io/solo-kit/blob/main/go.mod#L23) so that's what I've been using elsewhere as well

albertlockett (approved):

jenshu (approved):

PR #433, opened 2023-03-15
allow name override on operator without changing name of resources

josh-pritchard:
Allows me to generate templates in the new chart like this!
image

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/432


t-edris (commented):
Would it be worth it to make a test like [this one](https://github.com/solo-io/skv2/blob/master/codegen/cmd_test.go#L378) where we ensure that the `ValuesFileNameOverride` is used properly? Other than that, this looks like a simple enough change

josh-pritchard:
yep good call, just added a test

codegen/cmd_test.go
```diff
@@ -243,6 +243,94 @@ var _ = Describe("Cmd", func() {
 		Expect(envMapField.HeadComment).To(Equal("# Environment variables for the container. For more info, see the [Kubernetes\n# documentation](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#envvarsource-v1-core)."))
 	})
 
+	FIt("generates from templates using a name override", func() {
```
marcogschmidt, line 246: Unfocus

marcogschmidt (approved):
LGTM

t-edris (approved):

PR #442, opened 2023-04-04
fix input reconciler template

jenshu:
Fix compilation errors in input reconciler and input snapshot templates when running with k8s 1.25 libraries.

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/6833


conradhanson (approved):

nfuden (approved):
I dont see any other instances in the lib that could be problematic and given that this is ingested by both mesh and edge at this point if there are any gaps they really are dead code that should be removed

PR #445, opened 2023-05-15
Allow Deployment, SA, and Service to be applied to custom Operator namespace

jmhbh:
**Overview**
- The glooMeshPortalServer operator in gloo platform currently uses `{{ .Release.Namespace }}` which defaults to `gloo-mesh` we would like this to be `gloo-mesh-addons` instead. Therefore, by adding the `Namespace` field to the operator we can set this value and use it to control which namespace the deployment, service, and service account tied to the operator gets applied to.

**Testing**
- Tested this in gloo platform by running `make package-helm` and installing gloo platform locally.


fernandosanchezs (approved):
LGTM

samuelvl (approved):

PR #446, opened 2023-05-15
codegen: allow extra passthrough env vars

shashankram:
Allows setting passthrough env vars not explicitly defined in values.yaml. This can be used to set temporary env vars such as feature/experimental flags without needing to worry about breaking users as these values are not explicitly defined in the charts.

Example usage:
```
$ helm template install/helm/gloo-mesh-enterprise/ --set common.cluster=foo --set licenseKey=bar --set-string glooMeshMgmtServer.extraEnvs.FOO.value="true" | grep -A1 FOO
          - name: FOO
            value: "true"
```

```
$ helm template install/helm/gloo-mesh-enterprise/ --set common.cluster=foo --set licenseKey=bar --set glooMeshMgmtServer.extraEnvs.FOO.valueFrom.secretKeyRef.key=fookey --set glooMeshMgmtServer.extraEnvs.FOO.valueFrom.secretKeyRef.name=keyval --set glooMeshMgmtServer.extraEnvs.FOO.valueFrom.secretKeyRef.optional=true  | grep -A5 FOO
          - name: FOO
            valueFrom:
              secretKeyRef:
                key: fookey
                name: keyval
                optional: true
```


josh-pritchard:
Can we add some unit tests? Anything in the templating should have some basic tests. Something that similar to what your showing above

shashankram:
> Can we add some unit tests? Anything in the templating should have some basic tests. Something that similar to what your showing above

I thought the autogen files were enough. Let me look into adding something

shashankram:
@josh-pritchard added tests

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -95,6 +95,12 @@ spec:
 {{- if [[ $containerVar ]].env }}
         env:
 {{ toYaml [[ $containerVar ]].env | indent 10 }}
+{{- else if [[ $containerVar ]].extraEnvs }}
+        env:
+{{- end }}
+{{- range $name, $item := [[ $containerVar ]].extraEnvs }}
+          - name: {{ $name }}
+{{- $item | toYaml | nindent 12 }}
```
EItanya, line 103: Why is this list `else if` with `env`? Isn't this supposed to be an append style operation

shashankram: It is to ensure the `env:` top level key is always set if either of `env` or `extraEnvs` are set, but ensure `env:` doesn't get set twice if both are set.

EItanya (commented):
These changes LGTM. Would love @josh-pritchard  and @ilackarms  to take a look as well

EItanya (approved):

josh-pritchard (approved):
LGTM

PR #459, opened 2023-06-15
Make namespace configurable on operator

jmhbh:
**Overview**
- This [PR](https://github.com/solo-io/skv2/pull/445) added an optional option on the operator to set the namespace. I think the better option is to allow the namespace to be set in the custom values for an operator instead, and render the helm chart to use the custom value if it is set. 
- Net result in the helm chart is the following, and allows us to provide a default namespace for an operator while allowing it to be overridden as necessary
```
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: gloo-mesh-portal-server
    gloo-mesh: gloo-mesh-portal
  annotations:
    app.kubernetes.io/name: gloo-mesh-portal-server
  name: gloo-mesh-portal-server
  namespace: {{ default .Release.Namespace $.Values.glooPortalServer.namespace }}
spec:
...
```

**Testing**
- Tested this by installing gloo platform locally


albertlockett (dismissed):

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -50,8 +49,13 @@ metadata:
...
   name: [[ $operator.Name ]]
-  [[- if $customNamespace ]]
-  namespace: [[ $customNamespace ]]
+  [[- if $operator.Values ]]
+{{- $[[ $operatorVar ]] := [[ (opVar $operator) ]]}}
+{{- if $[[ $operatorVar ]].namespace }}
+  namespace: {{ $[[ $operatorVar ]].namespace }}
+{{- else }}
+  namespace: {{ .Release.Namespace }}
+{{- end }}
```
marcogschmidt, line 58: Nit: can this all be simplified using the `default` function? Not necessary but it makes the template a bit more concise. E.g. something like this

```
  namespace: {{ default .Release.Namespace [[ (opVar $operator) ]].namespace }}
```

jmhbh: Thanks for the comment, this is much cleaner and works well. Just tested it on a local install.

albertlockett (approved):

marcogschmidt (approved):

PR #463, opened 2023-07-17
Conditonal volumes, service and sidecar.

chunter0:


codegen/model/chart.go
```diff
@@ -88,6 +88,11 @@ type Operator struct {
 	// the name of the other operator can be included in this list. This operator
 	// will not be provisioned unless both are enabled (by having values.enabled = true)
 	EnabledDependsOn []string
+
+	// (Optional) Use this field to specify if this operator should be deployed as
+	// a sidecar of an existing operator. When this field is set, the RBAC rules
+	// will be merged as well as Deployment fields such as volumes and volume mounts.
+	Operator string
```
chunter0, line 95: ParentOperator

codegen/render/funcs.go
```diff
@@ -131,6 +131,22 @@ func makeTemplateFuncs(customFuncs template.FuncMap) template.FuncMap {
 		"containerConfigs": containerConfigs,
 
 		"opVar": opVar,
+
+		"operatorSidecars": func(operators []model.Operator) (out map[string][]model.Operator) {
+			for _, operator := range operators {
+				// Skip operators that specify a sidecar
```
chunter0, line 137: // Skip operators that dont specify a sidecar

codegen/cmd_test.go
```diff
@@ -40,6 +41,89 @@ var _ = Describe("Cmd", func() {
...
+						},
+						// Indicate this operator should be deployed as a sidecar
+						// of the glooMgmtServer operator
+						ParentOperator: "glooMgmtServer",
+					},
+				},
+			},
+		}
+
+		if err := cmd.Execute(); err != nil {
```
chunter0, line 121: I tested these changes via `go get` the commit directly into gloo-mesh-enterprise, I figure we can also have a unit test here but not sure where it should be placed

@josh-pritchard

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -81,9 +78,23 @@ spec:
 [[- if $volumes ]]
       volumes:
 [[ toYaml $volumes | indent 6 ]]
+[[- end ]]
+[[- range $container := $containers -]]
+[[- range $volume := $container.Volumes ]]
+[[- if (gt (len $container.EnableStatement) 0) ]]
```
ilackarms, line 84: nit: you could swap this line with th one above it, reads a bit cleaner imo

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -237,6 +251,40 @@ spec:
...
 
----
-
-[[ end ]]
\ No newline at end of file
+[[- range $container := $containers -]]
+[[- if and (gt (len $container.EnableStatement) 0) (gt (len $operator.Service.Ports) 0) ]]
+[[ printf "\n---" ]]
+[[ $containerVar := printf "$.Values.%s" (lower_camel $container.Name) ]]
+[[- if (gt (len $container.EnableStatement) 0) ]]
```
ilackarms, line 258: nit: we've already evaluated this to be true for this block

ilackarms (approved):

codegen/cmd_test.go
```diff
@@ -40,7 +41,103 @@ var _ = Describe("Cmd", func() {
...
+						CustomEnableCondition: `and ($.Values.glooAgent.enabled) (not $.Values.glooAgent.runAsSidecar)`,
+					},
+				},
+			},
+			ManifestRoot: "codegen/test/chart",
+		}
+
+		Expect(cmd.Execute()).NotTo(HaveOccurred(), "failed to execute command")
+
+		absPath, err := filepath.Abs("./test/chart/templates/deployment.yaml")
```
josh-pritchard, line 131: This path is used in another test, We are not really seeing these templates for review here as they are getting overwritten by a later test

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -237,6 +249,36 @@ spec:
...
+  type: {{ [[ $containerVar ]].serviceType }}
+  ports:
+  [[- range $port := $container.Service.Ports ]]
+  - name: [[ lower $port.Name ]]
+    port: {{ [[ $containerVar ]].ports.[[ $port.Name ]] }}
+  [[- end ]]
+{{- end }}
+[[- end ]]
+[[- end ]]
+[[ end ]]
```
josh-pritchard, line 284: This seems like a redundant Service, your test doesn't seem to reference the container.Service anyways so I am not sure what this is for, was this intended to be the sidecar?

josh-pritchard (commented):
This implementation seems to have changed a bit from when we originally chatted with @Sodman. Instead of using the values from a completely different operator now we appear to use the container sidecar config. I am okay with this but it does lead to some redundant configuration. I also think the values are now not being completely handled correctly with this approach as they will get generated for the glooAgent block and the sidecar but only the glooAgent block is referenced. Can I see an example of this run against GME in a draft PR? It's easy for me to confirm it’s all correct that way.

josh-pritchard (approved):
LGTM

ilackarms (approved):

PR #464, opened 2023-07-17
K8s 1.27 upgrade

sheidkamp:
Updates for k8s 1.27.

There is a bug we need to fix in `controller-manager` around how `nil` TerminalErrors are handled. The branch with that fix (forked off of the release branch) is here: https://github.com/solo-io/controller-runtime/tree/r0.15-terminal-error-nil-fix


changelog/v0.31.2/palceholder.yaml
```diff
@@ -0,0 +1,7 @@
+changelog:
```
jenshu, line 1: before we merge this we should be sure to add a valid changelog, and bump the minor version / make this a breaking change

sheidkamp: done

pkg/reconcile/v2/runner.go
```diff
@@ -146,7 +146,7 @@ func (r *runner[T]) RunReconciler(
 	}
 
 	// send us watch events
-	if err := ctl.Watch(&source.Kind{Type: obj}, &handler.EnqueueRequestForObject{}, predicates...); err != nil {
+	if err := ctl.Watch(source.Kind(r.mgr.GetCache(), obj), &handler.EnqueueRequestForObject{}, predicates...); err != nil {
```
nfuden, line 149: nice it mimicks upstream's change transparently! https://github.com/kubernetes-sigs/controller-runtime/commit/d6a053f8f0979d6f1528fe4d52b0d0f618352d77#diff-fa148e22912d4a8390270dd209fd4c81fad957622efe2d0dad8a5455d0589201R62

That being said are we sure that mgr is not nil?

pkg/source/dynamic.go
```diff
@@ -8,14 +8,13 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
-	"sigs.k8s.io/controller-runtime/pkg/runtime/inject"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 )
 
 // Stoppable is a stoppable source
 type Stoppable interface {
 	source.Source
-	inject.Stoppable
+	InjectStopChannel(<-chan struct{}) error
```
nfuden, line 17: Are we sure we should just be doing this update? It looks like the general path to this is no longer really supported upstream so I wonder if we need to do a larger change here.

pkg/handler/enqueue_static_handler.go
```diff
@@ -27,8 +29,11 @@ type BroadcastRequests struct {
 	WorkQueues *skqueue.MultiClusterQueues
 }
 
+// TODO: (sam-heilbron)
```
sam-heilbron, line 32: Let's remove this as we agreed it's not relevant

pkg/reconcile/v2/runner.go
sam-heilbron: I felt safe making this change, because earlier in this function there is:
```
gvk, err := apiutil.GVKForObject(obj, r.mgr.GetScheme())
```

pkg/handler/enqueue_static_handler.go
sheidkamp: done

pkg/reconcile/v2/runner.go
nfuden: yeah exactly

pkg/source/dynamic.go
sheidkamp: Seems like we don't have any reference to using the `Stoppable` in our repos: https://github.com/search?q=org%3Asolo-io%20Stoppable&type=code, and it is only used in this function signature: https://github.com/solo-io/skv2/blob/k8s-1.27-upgrade/pkg/source/dynamic.go#L99

Would it be appropriate to some kind of fail/panic to that `Add` function as an interim approach to 1) make sure none of the code we're upgrading as part of the k8s 1.27 is using it and 2) make sure if someone down the line tries to use it they get quick feedback instead of potentially having to debug a mysterious failure?

sheidkamp: change added below in `Add`. Nothing downstream seems to be having issues.

sheidkamp:
Work in progress until gloo-fed tests are passing

nfuden (approved):

PR #467, opened 2023-07-19
sort by key and description

puertomontt:


arianaw66 (approved):
Nit: I feel like Type should take priority over description

puertomontt:
AFAICT either the descriptions are equal and types differ or the types are equal and the descriptions differ.

jenshu (approved):
(i think we should fix the duplicate keys as mentioned in the linked issue, but this change LGTM)

PR #469, opened 2023-08-02
Fix portal helm installation

jmhbh:
- The gloo platform chart defines the `addonNamespace` as a common value [here](https://github.com/solo-io/gloo-mesh-enterprise/blob/main/codegen/helm/platform/platform_chart.go#L114). It is used to specify the namespace to install the addon components such as extauth as seen [here](https://github.com/solo-io/gloo-mesh-enterprise/blob/main/install/helm/gloo-platform/templates/ext-auth/deployment.yaml#L21). Since, the gloo portal server deployment yaml is generated by skv2, and we want to be able to use the same common value for the gloo portal server when installing the gloo platform helm chart, this PR introduces a new operator variable that allows us to specify the value path so we can render the following gloo portal deployment manifest.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app: gloo-mesh-portal-server
    gloo-mesh: gloo-mesh-portal
  annotations:
    app.kubernetes.io/name: gloo-mesh-portal-server
  name: gloo-mesh-portal-server
  namespace: {{ $.Values.common.addonNamespace | default $.Release.Namespace }}
spec:
...
```


marcogschmidt (commented):
Looks good! Would be good to have a unit test here, maybe in `cmd_test.go`. Can you check if that is something easy to add?

codegen/cmd_test.go
```diff
@@ -1188,6 +1188,119 @@ var _ = Describe("Cmd", func() {
...
+		}
+
+		err := cmd.Execute()
+		Expect(err).NotTo(HaveOccurred())
+
+		// Test that the deployment, SA, and CR templates are rendered correctly
+		deployment, err := os.ReadFile("codegen/test/chart/templates/deployment.yaml")
+		Expect(err).NotTo(HaveOccurred())
+
+		expectedDeploymentTmpl := "kind: Deployment\nmetadata:\n  labels:\n    app: painter\n  annotations:\n    app.kubernetes.io/name: painter\n  name: painter\n  namespace: {{ $.Values.common.namespace | default $.Release.Namespace }}"
```
marcogschmidt, line 1270: Nit: is it possible to use backticks and keep actual new lines? Would make it more readable

marcogschmidt (approved):

puertomontt (approved):

PR #478, opened 2023-08-14
[backport v0.31.x] Support Namespace-scoped RBAC

conradhanson:


ashleywang1 (approved):

shashankram:
I understand the change, but curious why this was backported to v0.31.x, because it isn't a backward compatible change. Ideally, we should not be breaking clients on patch releases.

shashankram:
> > I understand the change, but curious why this was backported to v0.31.x, because it isn't a backward compatible change. Ideally, we should not be breaking clients on patch releases.
> 
> Agreed we should not be breaking clients on patch releases.
> 
> gme uses v0.31.x and can't upgrade yet bc of k8s dependency upgrades so needed to backport change to this branch. For all intensive purposes, the breaking change is renaming the `Operator.Rbac` field to `Operator.ClusterRbac`. Technically, isn't backwards compatible, but functionally still the same.

We could have left the existing field and marked it as deprecated instead, and support both the old and new fields conditionally, if we really need to pull in such changes into patch releases.

PR #480, opened 2023-08-21
Support cluster or namespace scoping RBAC

conradhanson:


conradhanson:
no longer needed since gme on skv2 v0.34.x .

PR #481, opened 2023-08-21
create Roles & RoleBindings

conradhanson:
Adds roles and rolebindings to operator rbac template, where their generation is dictated by a helm flag called `namespacedRbac`. Template default when flag unset is to creates ClusterRole and ClusterRoleBinding.

Example helm api used in template is below:
```
gloo:
 namespacedRbac:
 - resources:
   - secrets
   - certs
   namespaces:
   - gloo-mesh
   - gloo-mesh-addons
```


ilackarms (approved):

codegen/test/chart/templates/rbac.yaml
```diff
@@ -6,43 +6,98 @@
...
+{{- if not (get $painterNamespacedResources secrets) }}
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - GET
+  - LIST
+  - WATCH
+{{- end }}
```
josh-pritchard, line 69: Where are there 2 cluster roles on this one?

josh-pritchard, line 69: Would this work if this condition isn't met and we create a ClusterRole with no rules?

codegen/test/chart/templates/rbac.yaml
```diff
@@ -6,43 +6,98 @@
...
+  {{- end }}
+{{- end }}
+
+{{- if not (eq (len $painterNamespacedResources) 1) }}
+---
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: painter
```
josh-pritchard, line 56: Where are there 2 cluster roles on this one where one has a <name>-namespace and one is just <name>

josh-pritchard (commented):
I think this mostly makes sense, still need to take a closer look. It would be very helpful to me if you could open a draft PR in GME so I could preview your indented usage over there, would you mind?

codegen/test/chart/templates/rbac.yaml
conradhanson: yes, you can create an empty clusterRole. I've confirmed it with a quick manual test on my local cluster.

We should avoid that scenario entirely though since we have an [if statement on line 50](https://github.com/solo-io/skv2/pull/481/files/1e0e3f7ff645d9dc622955c1cd47a4c57550f524#diff-66b811f6f6759f3ffc5585c2edac864224c006587e37951d28c7a4fe2be9b906) to prevent that.

codegen/test/chart/templates/rbac.yaml
conradhanson: The naming is so they don't conflict. 

We need both since the initial one is for resources that we don't support namespacing for (even if they are a namespace-scoped resource from a k8s perspective). We create the second with the eligible namespaced resources that aren't enabled from the user to be namespace-scoped.

josh-pritchard: We need to be careful about not having the release namespace in the name, ideally we would like certain components to be installed multiple times in different namespaces. If not today that should work eventually, by not having the release namespace in the cluster role the second install will always fail because the resource exists

conradhanson: We could add the release name to the second clusterrole name, and have the first clusterrole name include release name and namespace. That way it is still configured. Thiughts?

josh-pritchard (dismissed):
LGTM. Thanks Conrad, great changes

josh-pritchard (approved):

ilackarms (approved):

PR #483, opened 2023-08-22
custom ReadinessProbe

puertomontt:
this will allow passing templates for port


birkland (dismissed):

codegen/model/chart.go
```diff
@@ -133,6 +133,14 @@ type Container struct {
 	SecurityContext *corev1.SecurityContext
 }
 
+type ReadinessProbe struct {
+	Exec                []string // optional: if specified, the readiness probe will be an exec probe with the specified commands
+	Port                string
+	Path                string
```
ilackarms, line 139: nit: put `path` and `exec` next to each other and explain why only one should be used

birkland (approved):

Sodman (approved):
LGTM

Issue #488, opened 2023-08-30
Correctly render service port names with hyphens


shashankram:
ServicePort.Name with hyphens does not render correctly and will result in an error as follows:
`Error: parse error at (gloo-platform/templates/deployment.yaml:392): bad character U+002D '-'`

Allow hyphenated port names.

PR #489, opened 2023-08-30
codegen: correctly render port names with hyphen

shashankram:
Currently, port names with hyphens do not render correctly in the Helm template due how Helm processes hypenated values.

We would see an error as follows:
Error: parse error at (gloo-platform/templates/deployment.yaml:392): bad character U+002D '-'

The workaround is to use the Helm index function to correctly render service port names with hyphens.
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/488


codegen/test/chart/conditional-sidecar/templates/deployment.yaml
```diff
@@ -197,7 +197,7 @@ spec:
   type: {{ $glooMgmtServer.serviceType }}
   ports:
   - name: grpc
-    port: {{ $glooMgmtServer.ports.grpc }}
+    port: {{ index $glooMgmtServer "ports" "grpc" }}
```
lgadban, line 200: so this is the only existing test chart that has a `Service` with parameterized ports?

codegen/cmd_test.go
```diff
@@ -1804,6 +1802,81 @@ roleRef:
...
+				Expect(err).NotTo(HaveOccurred())
+				renderedSvc = &v1.Service{}
+				err = json.Unmarshal(bytes, renderedSvc)
+				Expect(err).NotTo(HaveOccurred())
+				break
+			}
+			Expect(renderedSvc).NotTo(BeNil())
+			Expect(renderedSvc.Spec.Ports[0].Name).To(Equal(portName))
+		},
+		Entry("port name without hyphen", "foo"),
```
lgadban, line 1876: might be overkill but is there value in testing that the value can be set by values (as well as the default value)?

codegen/test/chart/conditional-sidecar/templates/deployment.yaml
shashankram: Yes, otherwise they should have been updated with autogen.

codegen/cmd_test.go
shashankram: Don't think that's necessary because Helm allows overriding properties that have an hyphen, e.g. ` --set glooAgent.ports.grpc-runtimeapi=100`. We'd be testing Helm vs our templating, so seems like an overkill. I tested this manually though.

ilackarms (approved):

lgadban (approved):

codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -248,7 +248,7 @@ spec:
   ports:
   [[- range $port := $operator.Service.Ports ]]
   - name: [[ lower $port.Name ]]
-    port: {{ $[[ $operatorVar ]].ports.[[ $port.Name ]] }}
+    port: {{ index $[[ $operatorVar ]] "ports" "[[ $port.Name ]]" }}
```
chunter0, line 251: Note: This was missed for the extra service definition below for sidecar services

PR #498, opened 2023-09-07
Add svc port fix for sidecar services

chunter0:
https://github.com/solo-io/skv2/pull/489/files#r1318853643
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/488

### DIFF
--- a/changelog/v0.34.2/svc-port-name.yaml
+++ b/changelog/v0.34.2/svc-port-name.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: NEW_FEATURE
+    issueLink: https://github.com/solo-io/skv2/issues/488
+    description: >
+      Allow service port names to contain hyphens for sidecar services.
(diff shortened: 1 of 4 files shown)


lgadban (commented):
can we add a test for this similar to https://github.com/solo-io/skv2/blob/b3a7ab6a226d2689383dfc8acbbc314c31a7d519/codegen/cmd_test.go#L1879

chunter0:
> can we add a test for this similar to
> 
> https://github.com/solo-io/skv2/blob/b3a7ab6a226d2689383dfc8acbbc314c31a7d519/codegen/cmd_test.go#L1879

We have one: https://github.com/solo-io/skv2/pull/498/files#diff-fe4cd7f883cacb469abd837aa9d3421fd060356ed087c9d52d637056dc3190cdR155

chunter0:
Hmm I've pulled it into https://github.com/solo-io/gloo-mesh-enterprise/pull/10462 and test locally. But why write another test when its the same behavior inherited? @lgadban

lgadban:
Because the behavior is not inherited, it is a standalone change was made to the template. A test to enforce this behavior will prevent any regression in the future caused by e.g. unintentional changes to the template for sidecar service ports.

What is the concern with exercising this change in a test?

The framework is already in place by the linked test (https://github.com/solo-io/skv2/blob/b3a7ab6a226d2689383dfc8acbbc314c31a7d519/codegen/cmd_test.go#L1879). The test can either be augmented to have entries for the combinations of sidecar service ports or can be duplicated to explicitly test the sidecar service ports (vs. the "regular" service ports).

shashankram:
@chunter0 it makes sense to extend the existing service-port unit tests I added to test this scenario as well, as @lgadban has suggested. I would like to see this standalone test verify the service port name rendering correctly regardless of the usage in other tests as that could change and break this behavior.

lgadban (approved):

shashankram (approved):
Also backport

PR #500, opened 2023-09-14
[backport v0.29.x-gp] namespace-scoped rbac

conradhanson:
This backport includes api changes, which goes against semver (should be minor bump, but will be patch bump instead). Reasoning is that branch v0.29.x-gp is only used by gp v2.3.x, so we are breaking semver in order to support a backport to gme which requires these changes.


conradhanson:
[following this pr](https://github.com/solo-io/ext-auth-service/pull/634)'s footsteps for the changelog

/skip-changelog

ashleywang1 (approved):

PR #502, opened 2023-09-18
update operator rbac template resource naming to support multiple helm chart installations

conradhanson:


ilackarms (dismissed):

codegen/cmd_test.go
```diff
@@ -2062,7 +2062,7 @@ roleRef:
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: painter-{{ .Release.Name }}-{{ .Release.Namespace }}
+  name: painter-{{ .Release.Name }}-{{ .Release.Namespace }}-namespaced
```
samuelvl, line 2065: Can a `ClusterRole` be namespaced? Should we only add the `-namespaced` suffix to `Role` and `RoleBinding`?

EItanya: Namespaced refers to the usage of these resources, rather than the distinction between cluster wide and namespace specific

samuelvl: That makes sense!

samuelvl (approved):

PR #508, opened 2023-10-16
codegen: remove conditional enabling of internal CRDs

shashankram:
Internal CRDs are an implementation detail and must not be exposed to end users to enable/disable.

### DIFF
--- a/codegen/render/funcs.go
+++ b/codegen/render/funcs.go
@@ -33,6 +33,10 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const (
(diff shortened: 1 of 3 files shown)


codegen/render/funcs.go
```diff
@@ -33,6 +33,10 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+const (
+	internalCRDGroupName = "internal.gloo.solo.io"
+)
```
ilackarms, line 38: i would not hardcode this inside skv2

shashankram:
Will introduce feature specific conditions.

PR #510, opened 2023-10-16
codegen: allow templated env vars

shashankram:
Enables setting environment variables on containers
that can leverage templated Helm values.

A new field is needed for this because the existing
model.Env exposes the env vars in values.yaml which
isn't suitable when the env vars must use a templated
value based on other Helm values.

See https://github.com/solo-io/gloo-mesh-enterprise/pull/12570 for
example usage.


EItanya (approved):
This path forward makes sense to me

timflannagan (approved):

PR #516, opened 2023-11-22
Fix AppendingErrHandler issue causing errors not to be stored in passed down handlers

inFocus7:
# Changes

- Use a pointer receiver for `AppendErrorHandler`
- Add unit tests to make sure `AppendErrorHandler` works as expected (appending errors)
  - without the AppendingErrorHandler being a pointer receiver -> the tests will fail.
- Add very simple pull request template to add information regarding changes + context.

# Context

This was caught while debugging an issue in Portal. A failed `upsert` did not result in a `State: Failed` in Portal due to no errors in the `AppendingErrHandler`.

From more debugging I noticed that the `AppendingErrHandler` did have the expected error during `syncList` [here](https://github.com/solo-io/skv2/blob/3055431e9667ff6b93cac0b4ba5f64843c10566a/contrib/pkg/output/snapshot.go#L300-L306), but the error did not propagate upwards to `SyncLocalCluster` [here](https://github.com/solo-io/skv2/blob/3055431e9667ff6b93cac0b4ba5f64843c10566a/contrib/pkg/output/snapshot.go#L238).

This was due to reference vs value/copy modifications, where in this case the `errHandler` stored the error in a _copy_ of the object instead of the object itself. Updating the `AppendingErrHandler` methods to use `*AppendingErrHandler` fixed the issue.

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/517

### DIFF
--- a/contrib/pkg/output/errhandlers/err_handlers.go
+++ b/contrib/pkg/output/errhandlers/err_handlers.go
@@ -9,20 +9,30 @@ import (
 
 // this file contains ErrorHandlers for handling errors created when writing output snapshots
 
+func ResourceWriteError(resource ezkube.Object, err error) error {
+	return eris.Wrapf(err, "writing resource %v failed", sets.Key(resource))
+}
+func ResourceDeleteError(resource ezkube.Object, err error) error {
+	return eris.Wrapf(err, "deleting resource %v failed", sets.Key(resource))
+}
+func ListError(err error) error {
+	return eris.Wrapf(err, "listing failed")
+}
+
 type AppendingErrHandler struct {
 	errs error
 }
 
 func (a *AppendingErrHandler) HandleWriteError(resource ezkube.Object, err error) {
-	a.errs = multierror.Append(a.errs, eris.Wrapf(err, "writing resource %v failed", sets.Key(resource)))
+	a.errs = multierror.Append(a.errs, ResourceWriteError(resource, err))
 }
 
 func (a *AppendingErrHandler) HandleDeleteError(resource ezkube.Object, err error) {
-	a.errs = multierror.Append(a.errs, eris.Wrapf(err, "deleting resource %v failed", sets.Key(resource)))
+	a.errs = multierror.Append(a.errs, ResourceDeleteError(resource, err))
 }
 
 func (a *AppendingErrHandler) HandleListError(err error) {
-	a.errs = multierror.Append(a.errs, eris.Wrapf(err, "listing failed"))
+	a.errs = multierror.Append(a.errs, ListError(err))
 }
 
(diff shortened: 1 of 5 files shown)


contrib/pkg/output/errhandlers/err_handlers.go
```diff
@@ -9,23 +9,33 @@ import (
 
 // this file contains ErrorHandlers for handling errors created when writing output snapshots
 
+func ResourceWriteError(resource ezkube.Object, err error) error {
```
inFocus7, line 12: I separated the errors so I can easily `MatchError` in unit tests

contrib/pkg/output/errhandlers/err_handlers.go
```diff
@@ -9,23 +9,33 @@ import (
...
+	a.errs = multierror.Append(a.errs, ResourceDeleteError(resource, err))
 }
 
-func (a AppendingErrHandler) HandleListError(err error) {
-	a.errs = multierror.Append(a.errs, eris.Wrapf(err, "listing failed"))
+func (a *AppendingErrHandler) HandleListError(err error) {
+	a.errs = multierror.Append(a.errs, ListError(err))
 }
 
 // returns the errors collected by the handler
```
marcogschmidt, line 38: Nit: was already like this, but since we are here
```suggestion
// Errors returns the errors collected by the handler
```

contrib/pkg/output/errhandlers/err_handlers_test.go
```diff
@@ -0,0 +1,94 @@
...
+}
+func addDeleteError(handler output.ErrorHandler, resource ezkube.Object) {
+	handler.HandleDeleteError(resource, errors.New(""))
+}
+
+// errorExpectation is a function that expects a specific error to occur
+type errorExpectation func(err error, resource ezkube.Object)
+
+func expectListError(err error, resource ezkube.Object) {
+	Expect(err).To(HaveOccurred())
```
marcogschmidt, line 72: These should all be `ExpectWithOffset` since they are in a helper func.

marcogschmidt (commented):
LGTM, just a nit

contrib/pkg/output/errhandlers/err_handlers_test.go
inFocus7: Ah good catch! 👍🏼

marcogschmidt (dismissed):

jenshu (approved):

inFocus7:
github actions were unstable for a while and https://www.githubstatus.com states it's better now. going to look if I can kick them without having to push and require re-reviews

Issue #517, opened 2023-11-22
Fix AppendErrHandler not applying errors to original object


inFocus7:
While syncing, our error handlers maintain a list of errors that occurred during translations. There is an issue with the `AppendErrHandler` where the errors are only being appended to a copy of the handler, instead of the original handler object. This means that outer functions that have called 

(copied fro investigation on my end)

1. During resource translations, we sync the cluster using skv2 code, where we pass it an ErrorHandler. We expect this ErrorHandlers data to be modified with any errors catched. [[SyncLocalCluster reference](https://github.com/solo-io/skv2/blob/75d0d1f023e7fec1c2862eb18f3481c9954f61e0/contrib/pkg/output/snapshot.go#L226-L240)]
2. Inside SyncLocalCluster we call another function, syncList, and pass it the ErrorHandler.
3. syncList upserts the list of resources. [[syncList calling upsert reference](https://github.com/solo-io/skv2/blob/75d0d1f023e7fec1c2862eb18f3481c9954f61e0/contrib/pkg/output/snapshot.go#L300-L306)]
    a. **If any errors occur, we use the ErrorHandler and append to its list of errors.**
    b. **In syncList, after doing the above, the ErrorHandler object does have the error(s) we expect after writing it.**
4. We step out of `syncList` back to the `SyncLocalCluster` code.
5. **Now, in `SyncLocalCluster`, the ErrorHandler object does not have the error.**
6. [blah blah]
7. Now, the original function which called the cluster sync code does not have any error in the ErrorHandler it passed.

With the above steps in mind, we can see that between steps 3-5 there is a disconnect between the `ErrorHandler`’s errors in syncList (where we added the error), and the `ErrorHandler` in SyncLocalCluster because we don’t see the error there anymore.

Meaning there's some referencing funkiness occurring.

inFocus7:
Created an issue to have this more publicized and link it to my PR.

PR #520, opened 2023-11-29
Fix AppendingErrHandler issue causing errors not to be stored in passed down handlers

inFocus7:
_backport of #516_

# Description

Backporting onto a "patch" LTS branch, because upgrading Portal to skv2 latest requires k8s bumps + this is technically a breaking change and shouldn't be in a regular semver `patch` bump. 

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/517


inFocus7:
Cancelled workflows to not risk auto-merging before branch-protection rules are set.

PR #531, opened 2024-01-29
Remove extra  field so generated docs don't have duplicate  descripti…

chunter0:
If an operator has `enabled` field defined in the provided values `skv2` will generate another line for the default `enabled` field found when building helm chart values.

E.g:
```
type glooAgent struct{
  enabled bool `desc: "enabled gloo agent component"
}
```

Prior to this change would result in the following lines existing in the generated reference doc

```
|glooAgent.enabled|bool|enabled gloo agent component|false|
|glooAgent.enabled|bool|Enable creation of the deployment/service.|true|
```

BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/532

### DIFF
--- a/changelog/v0.36.4/remove-extra-field.yaml
+++ b/changelog/v0.36.4/remove-extra-field.yaml
@@ -1,0 +1,5 @@
+changelog:
+  - type: FIX
+    issueLink: https://github.com/solo-io/skv2/issues/532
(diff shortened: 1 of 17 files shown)


Rachael-Graham (approved):
LGTM on docs end

PR #535, opened 2024-03-08
Sai/merged sets

saiskee:
# Description

Please include a thorough summary of the changes.

This fixes ... \ This new feature can be used to ...

# Context

Summary of issue

Description and justification of any interesting decisions made

### DIFF
--- a/contrib/codegen/templates/sets/sets.gotmpl
+++ b/contrib/codegen/templates/sets/sets.gotmpl
@@ -344,6 +344,7 @@ func (s *{{ $kindLowerCamel }}MergedSet) Insert(
         {{ $kindLowerCamel }}List ...*{{ $import_prefix }}.{{ $resource.Kind }},
 ) {
     if s == nil {
+        panic("cannot insert into nil set")
     }
     if len(s.sets) == 0 {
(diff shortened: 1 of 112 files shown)


contrib/codegen/templates/sets/sets.gotmpl
```diff
@@ -240,7 +240,183 @@ func (s *{{ $kindLowerCamel }}Set) Clone() {{ $resource.Kind }}Set {
 	if s == nil {
 		return nil
 	}
-	return &{{ $kindLowerCamel }}Set{set: sksets.NewResourceSet(s.Generic().Clone().List()...)}
+	return &{{ $kindLowerCamel }}MergedSet{sets: []sksets.ResourceSet{s.Generic()}}
```
saiskee, line 243: This is the main effective change, clone now produces a merged set, which is a a list of sets which is iterated through on methods like `Find` or `List`

pkg/ezkube/resource_id.go
```diff
@@ -152,6 +153,37 @@ func KeyWithSeparator(id ResourceId, separator string) string {
 	return b.String()
 }
 
+func HashedKeyWithSeparator(id ResourceId, separator string) uint64 {
```
saiskee, line 156: Using a uint64 for the key may work here too

Issue #529, opened 2024-01-13
Need Gateway API TargetRef object


lgadban:
In order to expand our policies to attach to Gateway API resources (e.g. `Gateway`, `HTTPRoute`) we need to utilize the `TargetRef` API.

We should add this API in a reusable place for our projects, such as skv2

PR #536, opened 2024-03-08
rename TargetRefWithSectionName, add PolicyTargetReference

lgadban:
renaming message field but not used in the most recent LTS releases:
https://github.com/solo-io/gloo-mesh-enterprise/blob/v2.5.x/go.mod#L95
https://github.com/solo-io/gloo/blob/v1.16.x/go.mod#L56
BOT NOTES: 
resolves https://github.com/solo-io/skv2/issues/529

### DIFF
--- a/api/core/v1/core.proto
+++ b/api/core/v1/core.proto
@@ -162,14 +162,28 @@ message ObjectSelector {
     }
 }
 
-// TargetRef identifies Gateway API objects to apply a direct policy to.
+// PolicyTargetReference identifies Gateway API objects to directly apply policy to.
 // This is a copy of the upstream K8s Gateway API `targetRef` API.
-// Note that we are only using the upstream `PolicyTargetReferenceWithSectionName` type as currently our policies will distinguish between sections.
+// See the following for more information:
+// * https://gateway-api.sigs.k8s.io/geps/gep-713/
+// * https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyTargetReference
+// * https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L34-L56
+message PolicyTargetReference {
+  string group = 1;
+  string kind = 2;
+  string name = 3;
+
+  // Optional, if unspecified, the local namespace of the policy is inferred.
+  google.protobuf.StringValue namespace = 4;
+}
+
+// PolicyTargetReferenceWithSectionName identifies Gateway API objects, and optionally a specific section of those objects, to directly apply policy to.
+// This is a copy of the upstream K8s Gateway API `targetRef` API.
 // See the following for more information:
 // * https://gateway-api.sigs.k8s.io/geps/gep-713/
 // * https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyTargetReferenceWithSectionName
-// * https://github.com/kubernetes-sigs/gateway-api/blob/444631bfe06f3bcca5d0eadf1857eac1d369421d/apis/v1alpha2/policy_types.go#L58-L83
-message TargetRefWithSectionName {
+// * https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L58-L83
+message PolicyTargetReferenceWithSectionName {
   string group = 1;
   string kind = 2;
   string name = 3;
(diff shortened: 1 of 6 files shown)


Sodman (approved):
lgtm

api/core/v1/core.proto
```diff
@@ -162,14 +162,28 @@ message ObjectSelector {
...
 // This is a copy of the upstream K8s Gateway API `targetRef` API.
-// Note that we are only using the upstream `PolicyTargetReferenceWithSectionName` type as currently our policies will distinguish between sections.
 // See the following for more information:
 // * https://gateway-api.sigs.k8s.io/geps/gep-713/
 // * https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyTargetReferenceWithSectionName
-// * https://github.com/kubernetes-sigs/gateway-api/blob/444631bfe06f3bcca5d0eadf1857eac1d369421d/apis/v1alpha2/policy_types.go#L58-L83
-message TargetRefWithSectionName {
+// * https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L58-L83
+message PolicyTargetReferenceWithSectionName {
   string group = 1;
```
jenshu, line 187: should the fields in here be updated to mirror what's in https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io%2fv1alpha2.PolicyTargetReferenceWithSectionName?

api/core/v1/core.proto
```diff
@@ -162,14 +162,28 @@ message ObjectSelector {
...
+  string group = 1;
+  string kind = 2;
+  string name = 3;
+
+  // Optional, if unspecified, the local namespace of the policy is inferred.
+  google.protobuf.StringValue namespace = 4;
+}
+
+// PolicyTargetReferenceWithSectionName identifies Gateway API objects, and optionally a specific section of those objects, to directly apply policy to.
 // This is a copy of the upstream K8s Gateway API `targetRef` API.
```
jenshu, line 181: this comment is out of date, should be PolicyTargetReferenceWithSectionName API?

api/core/v1/core.proto
```diff
@@ -162,14 +162,28 @@ message ObjectSelector {
     }
 }
 
-// TargetRef identifies Gateway API objects to apply a direct policy to.
+// PolicyTargetReference identifies Gateway API objects to directly apply policy to.
+// This is a copy of the upstream K8s Gateway API `targetRef` API.
```
jenshu, line 166: this comment is out of date, should be PolicyTargetReference API?

changelog/v0.36.5/add-target-ref.yaml
```diff
@@ -0,0 +1,4 @@
+changelog:
```
jenshu, line 1: nit: since this is breaking, you could move to v0.37.0 and use type BREAKING_CHANGE

api/core/v1/core.proto
npolshakova: Yeah, I think the link is out of date, is it this one? https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L39

api/core/v1/core.proto
npolshakova: Yeah, can the PolicyTargetReference be inlined here like upstream is doing? https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L68

api/core/v1/core.proto
lgadban: the intent was the call out the generic `targetRef` concept not the specific message it is using.
I can update this though

api/core/v1/core.proto
lgadban: the intent was the call out the generic targetRef concept not the specific message it is using. will update though

lgadban: @npolshakova that link is the same as the one in the comment?
https://github.com/kubernetes-sigs/gateway-api/blob/b4f0307cc9269e73187300e72979e7e111ab74ab/apis/v1alpha2/policy_types.go#L34-L56

api/core/v1/core.proto
lgadban: you can not have anonymous nested objects in proto (as opposed to go struct)
have you seen a way to accomplish this?

changelog/v0.36.5/add-target-ref.yaml
lgadban: the goal was to fix this before it became a breaking change as no one has consumed this message yet (see PR description for references to edge & GME)
the versioning is a bit flimsy in this repo (e.g. there is currently not a `v0.36.x` brach) so it seemed easiest to just break it now while there are no consumers.

that being said, it is technically a break from a semver perspective, so I'm not sure the best approach

api/core/v1/core.proto
npolshakova: That link works- I must have miscopied it when trying to find it!

jenshu (approved):

PR #538, opened 2024-04-02
codegen: allow setting Strategy and PodSecurityContext

samuelvl:
### Description

Allow to define the following custom values in operator deployments:
- `deployment.spec.strategy` 
- `deployment.spec.template.spec.securityContext` 

### Context

This is required to enable persistence in Redis, allowing the utilization of the Recreate strategy and setting the fsGroup field, which is only available at the [pod-level security context](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#podsecuritycontext-v1-core).


codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -58,6 +60,16 @@ spec:
       [[- range $key, $value := $customPodLabels ]]
       [[ $key ]]: [[ $value ]]
       [[- end ]]
+[[- if and $deploymentStrategy ]]
```
puertomontt, line 63: why `and`?

samuelvl: I totally forgot to remove it

puertomontt (dismissed):

puertomontt (approved):

Nadine2016 (approved):

PR #549, opened 2024-04-22
codegen: support for conditional deployment strategy

samuelvl:
Allow configuring different deployment strategies based on specific Helm conditions:

```yaml
spec:
  strategy:
{{- if $.Values.painter.condition }}
    type: RollingUpdate
{{- end }}
{{- if not $.Values.painter.condition }}
    type: Recreate
{{- end }}
```

This is needed by Gloo Platform in order to use the `Recreate` strategy in the Redis deployment **only** when persistence is enabled.


codegen/templates/chart/operator-deployment.yamltmpl
```diff
@@ -60,14 +61,15 @@ spec:
...
-[[- if $deploymentStrategy.RollingUpdate ]]
-    rollingUpdate:
-[[ toYaml $deploymentStrategy.RollingUpdate | indent 6 ]]
+[[ toYaml $strategy | indent 4 ]]
+[[- else if $conditionalStrategy ]]
+  strategy:
+[[- range $s := $conditionalStrategy ]]
+{{- if [[ $s.Condition ]] }}
+[[ toYaml $s.Strategy | indent 4 ]]
+{{- end }}
```
chunter0, line 72: This seems like a bit much right? We just want to toggle the value of strategy here right? Couldn't we achieve that will something like this:

```yaml
strategy: {{ <user-value> (e.g: Recreate) | default <value> (e.g: RollingUpdate) }}
```

samuelvl: We want the strategy value to be automatically set based on a helm condition (if persistence is enabled or not), not based on user input. Something like this:

```yaml
spec:
  strategy:
{{- if $.Values.redis.deployment.persistence.enabled }}
    type: RollingUpdate
{{- end }}
{{- if not $.Values.redis.deployment.persistence.enabled }}
    type: Recreate
{{- end }}
```

The deployment strategy is an object, not a plain string https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#deploymentstrategy-v1-apps That's why we need the `toYaml` function

chunter0: This just seems a bit weird having both in the template. Could we not include the default case since kubernetes would handle that?

chunter0: Actually don't want to block you on this since its pretty small. But I'm good with the changes here. I'll give @EItanya the rest of today to approve otherwise I'll do it

chunter0 (approved):

k8so (approved):

PR #561, opened 2024-05-08
Sai/efficient union

saiskee:
# Description

This makes Union more efficient by checking if the input set is a sorted set, and taking advantage of merging two sorted lists (O(n+m))


contrib/pkg/sets/v2/sets.go
```diff
@@ -69,6 +69,10 @@ type ResourceSet[T client.Object] interface {
 	Clone() ResourceSet[T]
 	// ShallowCopy returns a shallow copy of the set
 	ShallowCopy() ResourceSet[T]
+
+	// Guarantees that when running Iter, Filter, or FilterOutAndCreateList, elements in the set will be processed in a
+	// sorted order by ResourceId. See ezkube.ResourceIdsCompare for the definition of ResourceId sorting.
+	IsSortedByResourceId() bool
```
EItanya, line 75: This feels very very specific and kind of strange to have on the interface to be honest

saiskee: We need a way to guarantee that the resource set in Union is sorted for more efficient merging. I'm open to other ideas on how we can do this.

conradhanson: I understand the need for knowing the set has been sorted by a specific type, but I don't like how this func is tied to implementation detail. I think making it generic with a name like `SortedBy` that uses/returns an enum value could make more sense. The enum would of course have to hold all known sorting types but this way it generalized, and the enum documents what all we sort by (yes that is implementation detail but abstracted).

EItanya: Don't these lists always have to be sorted? I actually think it's up to the caller to make sure they're sorted by the same thing. `Union` should just perform its operation.

saiskee: the `ResourceSet` interface does not guarantee that the data structure is sorted.

EItanya: Isn't the only implementation the list backed one at this point?

saiskee: Yes.. But leaving room for other implementations / sorting orders (creation timestamp, etc).

EItanya: Then those are all sorted

saiskee: Unless two resource sets are sorted by different criteria, then we'd have to check the sorting criteria are the same. If not, fallback to the receiving set's sort method?

saiskee: @EItanya is there a change you want to see here? not sure if refactoring this and figuring out the perfect semantics is a blocker here.

EItanya: You're right the semantics don't have to be perfect, but I think worrying about the sorting is exposing too many details. I would recommend removing this entirely and leaving it up to the caller.

EItanya (approved):

Sodman (approved):

PR #564, opened 2024-05-18
[Sets] Add Read lock to new `Union` and `ShallowCopy`

saiskee:


yuval-k (approved):

changelog/v0.40.1/sets-union-addRlock.yaml
```diff
@@ -0,0 +1,5 @@
+changelog:
+  - type: NON_USER_FACING
```
puertomontt, line 2: is this actually a fix?

puertomontt: does that mean release 0.40.0 should be marked as prerelease

puertomontt (approved):

PR #567, opened 2024-05-31
Add option to disable kubebuilder markers

arianaw66:
# Description

Adds option to disable kubebuilder markers and validations to omit them from the generated OpenAPI schema

# Context

This is needed for https://github.com/solo-io/gloo-mesh-enterprise/issues/17005

When disableKubeMarkers is set to true, `Type=object` annotations will be respected such that recursive fields will still work. `Required` annotations will also be respected.


arianaw66:
Yup, the GME repo changes are looking good!

marcogschmidt (approved):
LGTM

PR #573, opened 2024-07-15
Add conditional resources field

chunter0:
# Description

Support creating extra Kubernetes resources via skv2. The only thing here is we'd lose our current free for all templating, but I wanted to discuss as I don't believe we need the free-for-all behavior with our opinionated templates.

Do static resources like PersistentVolumeClaims need to be configurable from helm api perspective? I lean towards these types of resources being specific to how they fit in with the operators, otherwise I imagine for more flexibility a customer would bring their own version of said resource


codegen/cmd_test.go
```diff
@@ -41,6 +42,90 @@ var _ = Describe("Cmd", func() {
...
+									"app":       "gloo-mesh-redis",
+									"gloo-mesh": "gloo-mesh-redis",
+								},
+								Annotations: map[string]string{
+									"app.kubernetes.io/name": "gloo-mesh-redis",
+								},
+							},
+							Spec: corev1.PersistentVolumeClaimSpec{
+								AccessModes: []corev1.PersistentVolumeAccessMode{
+									corev1.ReadWriteOnce,
```
chunter0, line 82: So if we wanted to use this approach for defining extra resources related to a operator deployment. We wouldn't be able to template certain fields like [here](https://github.com/solo-io/gloo-mesh-enterprise/blob/7b2c7d5171ac9d41b36fc6cc6a2a8dcc167c804a/install/helm/gloo-platform/templates/redis-persistence.yaml#L15-L16). But then I asked myself do we really need to be allowing these fields to be configurable in the first place.

CC: @josh-pritchard @ilackarms

PR #576, opened 2024-08-07
Ignore specific markers in openapi generation from protos

jjamroga:
# Description
This allows for the user to define a list of kube markers to be ignored. Useful when referencing a .proto dependency which contains kube markers that aren't supported, and will never be supported.

Use case: I'm defining a proto that relies on some upstream dependency that also contains their own kube markers. These kube markers exist only for documentation, and do not affect the open api schema. Rather than implement noop markers to add to a marker registry, I would like to simply ignore them.

This adds:
- An additional argument to the protoc-gen-openapi executor

This changes:
- The way that protoc-gen-openapi arguments are structured (for readability's sake)

Related Issue: https://github.com/solo-io/protoc-gen-openapi/pull/27

# Testing Done
See parent PR


codegen/collector/executor.go
```diff
@@ -104,17 +108,16 @@ func (o *OpenApiProtocExecutor) Execute(protoFile string, toFile string, imports
 	_, fileName := filepath.Split(protoFile)
 	directoryName := fileName[0 : len(fileName)-len(filepath.Ext(fileName))]
 
+	baseArgument := "--openapi_out=yaml=true,single_file=false,include_description=true,multiline_description=true,proto_oneof=true,int_native=true"
+	baseArgument = fmt.Sprintf("%s,enum_as_int_or_string=%v", baseArgument, o.EnumAsIntOrString)
+	baseArgument = fmt.Sprintf("%s,additional_empty_schema=%v", baseArgument, strings.Join(o.MessagesWithEmptySchema, "+"))
+	baseArgument = fmt.Sprintf("%s,disable_kube_markers=%v", baseArgument, o.DisableKubeMarkers)
+	baseArgument = fmt.Sprintf("%s,ignored_kube_marker_substrings=%v", baseArgument, strings.Join(o.IgnoredKubeMarkerSubstrings, "+"))
```
shashankram, line 115: If you want to make this more readable, I'd group these options into a slice and use `strings.Join()`.

shashankram (approved):

codegen/collector/executor.go
jjamroga: Yeah honestly I should have done that. I think this is better than what we have, so if we revisit this and add another argument we could improve this again.

k8so (approved):

PR #579, opened 2024-08-22
Default release namespace for namespaced rbac

ashleywang1:
# Description

In https://github.com/solo-io/gloo-mesh-enterprise/pull/17198, we added Pods, ConfigMaps, and Secrets to the DashboardInputSnapshot, which caused the gloo-mesh-ui to begin reading all of these resources from all namespaces in the mgmt-cluster. We intended to only allow these resources to be read from the 'gloo-mesh' namespace, but the NamespacedRbac requires you to pass in glooUi.namespacedRbac values in order to activate that feature.

By default, we want to set the glooUi.namespacedRbac.resources values, and no glooUi.namespacedRbac.namespaces value. Then we want to make a skv2 change so that if glooUi.namespacedRbac.namespaces is empty but glooUi.namespacedRbac.resources is not, we default to the Release.Namespace.
BOT NOTES: 
resolves https://github.com/solo-io/gloo-mesh-enterprise/issues/15739


conradhanson:
Instead of altering these templates, why don't we just set defaults in GME's gloo-platform helm chart? Default being the 3 resources we need along with the release namespace for glooUi.namespacedRbac.

ashleywang1:
> Instead of altering these templates, why don't we just set defaults in GME's gloo-platform helm chart? Default being the 3 resources we need along with the release namespace for glooUi.namespacedRbac.

We can't set the release namespace by default in the gloo-platform helm chart - see https://github.com/solo-io/gloo-mesh-enterprise/issues/15739#issuecomment-2313169671

jehawley (dismissed):

conradhanson (approved):
lgtm!

josh-pritchard (approved):

PR #584, opened 2024-09-09
Kubernetes 1.31 and go1.23

tjons:
# Description

This PR upgrades our kubernetes dependencies to v1.31, and upgrades our golang version to 1.23.1. 

# Context

The `"k8s.io/client-go/util/workqueue"` package has deprecated the `workqueue.RateLimiterInterface` type in favor of a generic type, so that is the bulk of the code changes. The changes here (https://github.com/solo-io/skv2/pull/584/files#diff-db74ca5e769dd8b6314512f2a0ffb0532831ac381c2dcef8c3924cee208f9f63R42) are due to a new package level register for globally unique controllers that does not reset when a controller is removed. To allow our tests to run as desired, we disable unique name validation in the test suite. 

BOT NOTES: 
resolves https://github.com/solo-io/gloo/issues/9683


puertomontt (dismissed):

test/test_config.go
```diff
@@ -37,6 +39,13 @@ func MustManager(ctx context.Context, ns string) manager.Manager {
 		Cache: cache.Options{
 			DefaultNamespaces: map[string]cache.Config{ns: cache.Config{}},
 		},
+		Controller: config.Controller{
```
sam-heilbron, line 42: I could ❤️ this comment any more. Super useful to have the c-r issue, and the rationale for why we skip it

sam-heilbron (approved):

puertomontt (approved):

PR #595, opened 2025-04-15
Minify Helm chart values JSON schema

marcogschmidt:
## Motivation
With this change, the `values.schema.json` file generated as part of the Helm chart generation is no longer prettified. This drastically reduces the size of the file and minimizes the changes of the new maximum file size limit [introduced by Helm](https://github.com/helm/helm/commit/e4da49785aa6e6ee2b86efd5dd9e43400318262b) in `v3.17.3`. There are no functional changes. Human users can still prettify the file using external tools (e.g. `jq`) if necessary.

## Testing
You can test that this fixes the GME issue by following the instructions on https://github.com/solo-io/gloo-mesh-enterprise/pull/20443.


codegen/render/funcs.go
```diff
@@ -693,7 +658,7 @@ func toJSONSchema(values values.UserHelmValues) string {
 		panic(err)
 	}
 
-	return indentJson(string(jsonSchema))
+	return string(jsonSchema)
```
marcogschmidt, line 661: This is the only change.

jenshu (approved):

birkland:
Would it be possible to bump helm in CI to v3.17.3?  That would difinitively prove GME can be installed with helm v3.17.3 with the minified json.

Also, while you're at it, could you bump the helm dependency in go.mod to `helm.sh/helm/v3 v3.17.3`?  It's on `v3.17.0`, and all versions less than `v3.17.3` will register as a CVE

codegen/render/funcs_test.go
```diff
@@ -26,6 +28,49 @@ func prepareExpected(expected string) string {
 	return expected
 }
 
+// minifyJSON is a helper function for tests to remove whitespace from JSON strings
+func minifyJSON(jsonStr string) (string, error) {
```
jenshu, line 32: json.Compact might work here https://pkg.go.dev/encoding/json#Compact

Sodman (commented):
LGTM although I agree with Aaron's feedback that we should bump ourselves out of CVEs with this change too. Also agree with Jenny that if there's a stdlib way to do the same thing, I'd prefer that over a DIY solution.

codegen/render/funcs_test.go
marcogschmidt: Thanks for the pointer, I wasn't aware of that function! Just tested and it works like a charm 🎉

marcogschmidt:
> Would it be possible to bump helm in CI to v3.17.3? That would difinitively prove GME can be installed with helm v3.17.3 with the minified json.
> 
> Also, while you're at it, could you bump the helm dependency in go.mod to `helm.sh/helm/v3 v3.17.3`? It's on `v3.17.0`, and all versions less than `v3.17.3` will register as a CVE

These considerations totally make sense and I was planning on bumping the dependency, but we need to do that in the GME repo. We don't have any Go dependency on helm here. Did you perchance think this PR is in the GME repo?

birkland (approved):

Sodman (approved):

